PR #14, opened 2025-02-27
refactor(yarn/node): try to upgrade outdated yarn/node dependencies

Madman10K:
This should upgrade all our dependencies and should fix all our security issues. Also added a dependabot config file


node-packages/yarn-project.nix
```diff
@@ -52,9 +52,7 @@ let
       rm $out/.gitignore
     '';
     outputHashMode = "recursive";
-    outputHash = (if stdenv.isAarch64
```
alehander92, line 55: we have to keep this `if`: we have two different hashes for mac and linux IIRC; i think you can get the hash for mac, on linux, by running something like `nix build --dry-run .#packages.aarch64-darwin.something` (as @PetarKirov has suggested before)

we can also make this `if` somehow auto-updatable by the plugin, but not in the scope of this PR

Madman10K: I didn't know this was a thing. Thanks

PR #20, opened 2025-03-01
Refactor public folder(Rebase and update from gitlab)

Madman10K:
Refactors the public folder so that it is divided into 2 directories:

1. `resources` - this is for our own code/resources/fonts
1. `third_party`
   - `third_party/*/` - every library in a folder is a third-party library
   - `third_party/*` - every file in the root `third_party` directory is either a single-file third-party library or a modified wrapper on top of a given third-party library

Testing: tested locally with Tup and with the Nix package. No testing has been done using the appimage scripts or the non-nix build shell scripts, though it should work fine as is.


alehander92 (approved):
seems to work locally!

PR #54, opened 2025-03-17
feat: Change notification style and placement

pxor:
Apply redesign for the notifications

Main notifications:
Screenshot from 2025-03-17 14-08-47

Secondary notifications:
Screenshot from 2025-03-17 14-08-53


alehander92 (commented):
good initial implementation of the updated design, nice!

alehander92 (approved):

PR #61, opened 2025-03-19
fix: Install subwindow checkbox :after element

pxor:
Minor fix for the checkbox and set the traceSharingEnabled to false
image

### DIFF
--- a/src/config/default_config.yaml
+++ b/src/config/default_config.yaml
@@ -10,7 +10,7 @@ events: true
 # history: experimental support
 history: true
 repl: true
-traceSharingEnabled: true
+traceSharingEnabled: false
 
 # === feature settings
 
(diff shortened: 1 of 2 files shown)


src/config/default_config.yaml
```diff
@@ -10,7 +10,7 @@ events: true
 # history: experimental support
 history: true
 repl: true
-traceSharingEnabled: true
+traceSharingEnabled: false
```
alehander92, line 13: Please, update it in both copies of the config file: config/ and src/config/. we need to move all usages to src/config but this is not part of this PR

alehander92 (changes requested):

alehander92 (approved):

src/config/default_config.yaml
alehander92: (i was told it's already `false` there: sorry)

PR #62, opened 2025-03-19
fix-mac-build

SersemPeca:
This PR downgrades the SDK with which we build CodeTracer for mac. It also adds the build_dmg.sh script that packages the CodeTracer.app into a dmg with dmg_background.png as its mount background


alehander92 (approved):
look good: i can't test on mac, so I assume the test on Nedy's and Niki's macs is good: i'll wait for a build/or confirmation before closing the issue though

PR #72, opened 2025-03-20
Fix: experiments, optimizations and refactoring for the ruby recorder

alehander92:
experimented with recording rubocop, however not so many optimizatations for now. We do filter out steps to `gems/` now, but more parametrization/fixes are still remaining


alehander92:
depending on https://github.com/metacraft-labs/codetracer-ruby-recorder/pull/4 working for small examples too

PR #73, opened 2025-03-20
fix: Handle https urls with sslContext; compile ct with  `-d:ssl` in dev build/packages

pxor:


alehander92 (commented):
good: in the future, if wee need , we can change to `CVerifyPeer` if the domain has valid certificates, but this would be an easier fix.

src/Tuprules.tup
```diff
@@ -3,6 +3,7 @@ NIM_COMMON_FLAGS=\
 -d:chronicles_sinks=json\
 -d:chronicles_line_numbers=true\
 -d:chronicles_timestamps=UnixTime\
+-d:ssl\
```
alehander92, line 6: please, try to add now this flag to the other `ct` builds: the nix build, the appimage one and the mac one

alehander92 (approved):
merging to include package fixes (by @SersemPeca & @pxor coop) and SSL support

TODO in further PR-s address review/concerns of Nikola: 
* refactor files/code for trace sharing
* use streaming to support bigger traces

PR #85, opened 2025-03-26
fix(appimage): Downngrade ssl dependency; Start using some more of it

SersemPeca:


nix/packages/default.nix
```diff
@@ -31,6 +31,8 @@
 
         openssl = pkgs.openssl;
 
+        openssl_1_1 = pkgs.openssl_1_1;
```
alehander92, line 34: is this used: do we need both `openssl` and `openssl_1_1` here?

alehander92:
(we should just use `openssl` passing such a flag to nim as well, however currently work is focused on wasm/stylus support: eventually merge this after an `openssl3` fix )

alehander92:
superseded by #126 : closing

PR #86, opened 2025-03-31
fix: Handle new changes of upload/download api

pxor:


alehander92 (approved):
Nikola and Nedy improved the streaming and overall logic, and Nikola added a few tests; Nedy addressed several small things by me (a small trace path lang detection bug and nitpicks) 

There are some bugs with opening from download link from the startup menu where it seems the content from event log/filesystem disappears, or that the editor doesn't open: but those are a bit flaky, they don't easily reproduce. I am not sure if they're new or a regression, but I assume they might be realted to some slowness/timing, not so much to the new logic(maybe if it's a bit slower/heavier sometimes to that, but they're probably more fundamental problems): approved!

PR #102, opened 2025-04-17
feat: Add progress bar for uploading process

pxor:
Add initial support for progress bar, still needs more work to be done

image


pxor:
After minor cosmetic changes

image

src/ct/online_sharing/online_sharing_test.nim
```diff
@@ -1 +1,29 @@
...
+
+    echo "\nDownloading"
+    let (fileId, password) = extractInfoFromKey(info.downloadKey, conf)
+    let newId = downloadTrace(fileId, info.downloadKey, password, conf)
+    echo "Downloaded trace ID: ", newId
+    check newId >= 0
+
+    echo "Deleting"
+    deleteRemoteFile(traceId, info.controlId, conf)
+    expect Exception:
```
alehander92, line 28: does this compile? no '`try` i think

src/frontend/lib.nim
```diff
@@ -421,6 +421,34 @@ when defined(ctIndex) or defined(ctTest):
     # for more info, please read the comment for `setupLdLibraryPath` there
     nodeProcess.env["LD_LIBRARY_PATH"] = nodeProcess.env["CT_LD_LIBRARY_PATH"]
 
+  proc runUploadWithStreaming*(
+      path: cstring,
+      args: seq[cstring],
+      onData: proc(data: string),
```
alehander92, line 427: we could use `cstring` here, but it can work like that as well, maybe with more than needed string conversions

src/ct/online_sharing/online_sharing_test.nim
alehander92: @pxor told me this is `expect`, not `except`, sorry !

alehander92 (approved):
good work: chatted a bit with Nedy, he explained more things, it seems some temp files might be left in trace folders (maybe when interrupting upload?), but not a blocker; also some other small things, but seems good to merge

(note: zip progress tracking seems to be based on count of files, not chunks, for ram/resource reasons; this is also useful, but if one has e.g. a single huge file and some smaller, it's not very helpful: not a blocker)

PR #104, opened 2025-04-24
Fix: Ruby omniscience

pxor:
Working on integrating the omniscience with the Ruby language. This branch allows for loops with 'while' and 'until' to be integrated and top-level condition statements to be seen properly


Franz-Fischbach (approved):
If branches and omniscience loop controls work now.
Tested using examples/ruby_space_ship

Madman10K (approved):
LGTM! Looks like a sensible solution

PR #112, opened 2025-05-02
fix: Download and import Db trace properly

pxor:
This PR works on fixing remote sharing from other computers as we had some assumptions in the importDbTrace function
And fix the wasm language detection on download


alehander92:
OK: merged in `feat-wasm-tracing` , which is still PR #113

PR #113, opened 2025-05-05
Support/fixes for wasm tracing (and download/import db trace fixes by Nedi)

alehander92:
also adapt to runtime_tracing 0.10.0 with a temp branch of our older noir commit that we were supporting(adapting also to runtime_tracing 0.10.0: TODO in the future, not this PR: eventually integrate with newer noir which initiated actual support for those values)

rebased on `feat-upgrade-db-backend-runtime-tracing` from https://github.com/metacraft-labs/codetracer/pull/95 : closing #95 as it's integrated here

### DIFF
--- a/src/common/common_types.nim
+++ b/src/common/common_types.nim
@@ -844,8 +844,10 @@ type
   EventLogKind* {.pure.} = enum ## EventLog kinds
     Write,
     WriteFile,
+    WriteOther,
     Read,
     ReadFile,
+    ReadOther,
     # not really used for now
     # we might remove them or implement them
     # in the future
@@ -867,7 +869,9 @@ type
     highLevelLine*: int
     # eventually: might be available in the future
     # lowLevelLocation*: Location
-    filenameMetadata*: langstring ## metadata for read/write file events:
+    ## metadata: file for read/write file events 
+    ## or a more general kind of key for WriteOther/ReadOther
+    metadata*: langstring 
     bytes*: int
     stdout*: bool
     directLocationRRTicks*: int
@@ -951,7 +955,7 @@ type
     lowLevelLocation*: langstring
     kind*: EventLogKind
     content*: langstring
-    filenameMetadata*: langstring
+    metadata*: langstring
     base64Encoded*: bool
     stdout*: bool
 
--- a/src/common/common_types.nim
+++ b/src/common/common_types.nim
@@ -416,6 +416,7 @@ type
     member*: seq[Value]
     refValue*: Value
     address*: langstring
+    isMutable*: bool
     strong*: int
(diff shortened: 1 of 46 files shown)


src/ct/trace/record.nim
```diff
@@ -1,20 +1,28 @@
-import std/[os, osproc, strutils, sequtils],
+import std/[os, osproc, strutils, sequtils, strtabs, tables],
   ../../common/[ lang, paths, types, trace_index, config ],
   ../utilities/[language_detection ],
   ../cli/build
 
-proc recordInternal(exe: string, args: seq[string], config_path: string): Trace =
-  var env = setup_env(configPath)
+proc recordInternal(exe: string, args: seq[string], configPath: string): Trace =
```
SersemPeca, line 6: What is passed along for configPath when using Codetracer in a non-dev build ?

alehander92: currently an empty string, and in this case we just try to copy and pass the existing env

SersemPeca (approved):
LGTM

src/db-backend/src/tracepoint_interpreter/executor.rs
```diff
@@ -188,7 +188,17 @@ pub fn execute_bytecode(
                     #[allow(clippy::unwrap_used)]
                     let array = stack.pop().unwrap();
 
-                    if let ValueRecord::Sequence { elements, type_id: _ } = array {
+                    if let ValueRecord::Sequence { elements, type_id: _ , is_slice} = array {
+                        if is_slice {
+                            let mut err_value = Value::new(TypeKind::Error, eval_error_type.clone());
+                            err_value.msg = "Slices not supported currently".to_string();
```
Ro6afF, line 194: Why is this error? Aren't slice records the same as normal sequence records?

src/common/common_types.nim
```diff
@@ -416,6 +416,7 @@ type
     member*: seq[Value]
     refValue*: Value
     address*: langstring
+    isMutable*: bool
```
Ro6afF, line 419: Should we add the other fields, introduced in the new version of `runtime_tracing`?

alehander92: metadata should be there, but i don't think there are others: or their conversion is still TODO e.g. like for slices

src/db-backend/src/tracepoint_interpreter/executor.rs
alehander92: good question: not sure, maybe i supposed that slices have different internal representation(pointer/length?), or that it's not obvious if it's safe to just pass them around in the same way: can't really remember

PR #122, opened 2025-05-07
Automate releases in CI

Madman10K:
This PR adds an automation that automatically pushes a draft release when a tag is pushed. Adds a changelog text from `CHANGELOG.md` and adds a resource tarball for use by our Linux desktop packages.


alehander92 (approved):
looks good!

PR #123, opened 2025-05-08
VsCode Extension support + DAP + frontend reform (ongoing work)

pxor:
(alexander later EDIT):  
* VsCode Extension support (the vscode extension itself is separate! but uses parts/code built from the current `codetracer` repo: here https://github.com/metacraft-labs/codetracer-vscode-extension/ )
* a frontend reform: a new mediator pattern-based way to communicate for components; making components  more self-contained/reusable; overrides of transport for extension/single page cases (so we can support both extension webviews and the classic electron/browser single page case)
* adapting ui components/code for the reform;
* use DAP instead of our custom protocol for the db backend
* eventually: simplifying index.nim and moving a lot of that to the backend itself (maybe incrementally/if we have the resources)

for now new work on this might continue in feature branches and PR-s to this branch;  it currently breaks the electron usecase, so we can't still merge it to the main branch!

EDIT: now most of the electron usecase should be restored, with IMO a few small regressions (maybe some notifications, scratchpad focus/reopen if closed on `add to scratchpad`(EDIT: this might be fixed in #334 ), maybe more bugs?)


Ro6afF:
Oops, commited to the wrong branch. I removed the commits. Sorry.

PR #124, opened 2025-05-12
Refactor config style

Madman10K:
Changes in this PR:

- Removed the duplicate `config` folder
- Refactored all files to use `src/config` instead of `config`
- Moved the YAML in `default_config.yaml` to kebab-case
- Moved flow settings under a `flow` object
- Moved trace sharing settings under a `trace-sharing` object
- Renamed `map` to `bindings`
- Renamed `v` to `version`

TODO:

- Add macOS key bindings


src/frontend/index_config.nim
```diff
@@ -903,8 +904,8 @@ proc initDebugger*(main: js, trace: Trace, config: Config, helpers: Helpers) {.a
         paths: trace.sourceFolders,
         traceID: trace.id,
         calltrace: config.calltrace and trace.calltrace,
-        preloadEnabled: config.flow and trace.lang != LangPython,
-        callArgsEnabled: config.callArgs,
+        preloadEnabled: config.flow.enabled and trace.lang != LangPython,
+        callArgsEnabled: config.toJs["call-args"].to(bool), # Does not convert correctly with config.`call-args`?? Nim, what the fuck!?
```
alehander92: the frustration is understandable, but please, don't use this kind of language

alehander92 (changes requested):
overally looks very good, a nice improvement!

Using the native language variable case for `Config` fields seems as the best solution, independently of the lisp-case in the config files: i wasn't sure, but the source code and some of the bugs we had seems convincing: it's more natural to use the language-specific camelCase for nim(or if eventually we need something like that: underline_case in rust)

alehander92:
PS: Important: one must fix the nix, macos and appimage builds as part of this PR if one wants to fix the source config issue(storing config only in `src/config`, not in `config/`): sorry for that

Madman10K:
Changing to kebab case while keeping the original case of the source code is too much effort. Maybe I will try again when we move to nim 2.

The move to separate config objects and the removal of the top level config directory still remains.

alehander92 (approved):
good work!

PR #125, opened 2025-05-15
Add dependabot config for github action workflows

Madman10K:
Some of our workflows are sensitive to security vulnerabilities, so it's good to have dependabot set up to remind us to update these workflows as soon as possible. This config is needed because there are no security notifications about action workflows in the security tab as of now, so this is the only way to get notified about such issues.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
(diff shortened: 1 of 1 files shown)


alehander92 (commented):
it looks reasonable: this will send us a notification/review once a month? or a PR?

Madman10K:
@alehander92 This will check for an update once a month. If there is a need to upgrade something a PR will be created for it automatically

Ro6afF (approved):

PR #126, opened 2025-05-19
fix(ssl): Use ssl3 for ct binary

SersemPeca:
TODO: The versions of the ssl package in nixpkgs DO NOT coincide with standard ssl releases.
We should figure out a way to get the correct libraries we need

### DIFF
--- a/appimage-scripts/build_appimage.sh
+++ b/appimage-scripts/build_appimage.sh
@@ -76,6 +76,8 @@ cp -L "${LIBZIP}"/lib/libzip.so.5 "${APP_DIR}"/lib
 
 OPENSSL=$(nix eval --raw "${ROOT_PATH}#packages.${CURRENT_NIX_SYSTEM}.openssl.out")
 cp -L "${OPENSSL}"/lib/libssl.so.3 "${APP_DIR}"/lib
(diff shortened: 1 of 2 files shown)


appimage-scripts/build_appimage.sh
```diff
@@ -15,7 +15,7 @@ cleanup() {
   rm -rf ./squashfs-root
 }
 
-trap cleanup EXIT
+# trap cleanup EXIT
```
alehander92: should it be commented out?

appimage-scripts/result-bin
```diff
@@ -0,0 +1 @@
+/nix/store/lygl27c44xv73kx1spskcgvzwq7z337c-openssl-3.3.2-bin
```
alehander92: should those be committed?

alehander92:
this is a new, improved version of https://github.com/metacraft-labs/codetracer/pull/85, right? so we can close #85  i assume

appimage-scripts/result-bin
alehander92: (ok, removed)

appimage-scripts/build_appimage.sh
alehander92: restoring it

PR #127, opened 2025-05-19
fix `ct run` for now for wasm/rust (but not other wasm langs)

alehander92:
cleanup/comment out/remove some debug prints


lispegistus (approved):
LGTM. Preferably remove commented out code.

Issue #128, opened 2025-05-19
CodeTracer fails to launch due to bad temp location


Madman10K:
After running `just build-macos-app` and installing codetracer the resulting binary does not start.

After running it manually:

```
dlopen(libcrypto.3.dylib, 0x0002): tried: 'libcrypto.3.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OSlibcrypto.3.dylib' (no such file), '/usr/lib/libcrypto.3.dylib' (no such file, not in dyld cache), 'libcrypto.3.dylib' (no such file)
dlopen(libcrypto.1.1.dylib, 0x0002): tried: 'libcrypto.1.1.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OSlibcrypto.1.1.dylib' (no such file), '/usr/lib/libcrypto.1.1.dylib' (no such file, not in dyld cache), 'libcrypto.1.1.dylib' (no such file)
dlopen(libcrypto.38.dylib, 0x0002): tried: 'libcrypto.38.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OSlibcrypto.38.dylib' (no such file), '/usr/lib/libcrypto.38.dylib' (no such file, not in dyld cache), 'libcrypto.38.dylib' (no such file)
dlopen(libcrypto.39.dylib, 0x0002): tried: 'libcrypto.39.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OSlibcrypto.39.dylib' (no such file), '/usr/lib/libcrypto.39.dylib' (no such file, not in dyld cache), 'libcrypto.39.dylib' (no such file)
dlopen(libssl.3.dylib, 0x0002): tried: 'libssl.3.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OSlibssl.3.dylib' (no such file), '/usr/lib/libssl.3.dylib' (no such file, not in dyld cache), 'libssl.3.dylib' (no such file)
dlopen(libssl.1.1.dylib, 0x0002): tried: 'libssl.1.1.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OSlibssl.1.1.dylib' (no such file), '/usr/lib/libssl.1.1.dylib' (no such file, not in dyld cache), 'libssl.1.1.dylib' (no such file)
[...]

PR #129, opened 2025-05-20
Refactor the way we interact with temporary files

Madman10K:
# DO NOT MERGE YET
This works well on Linux, but it's best if I test it on macOS as well before merging it.

## Content
This refactor should create a single source of truth about all temporary files. The new added conditional logic for macOS also fixes the bug in #128.

Misc things in this PR:

1. Fixed some Nim compiler warnings
1. Fixed 1 Nim deprecation warning
1. Removed some dead code and dead files along the way

> [!WARNING]
> Some references to `/tmp` still exist in some files such as:
> 1. `src/tools/log_task.py` - These might not be needed?
> 1. `src/tester/tester.nim` and `src/tester/parallel_tester.nim` - tester should be deprecated by now
> 1. `src/Tuprules.tup` - No need to modify developer-specific files, especially when there's no sandboxing for `tup` anyway so there will be no effect here.

## Breaking changes
Due to macOS' issue with the `/tmp` directory, I am forced to write to set my temp directory to `$TMP/codetracer/`. This means that temporary files for codetracer are now under `$TMP/codetracer/codetracer` instead. Of course, this makes navigating to different runs a bit more cumbersome, though the structure of our temporary folder is cleaner this way + it's better to have 1 unified directory structure instead of juggling 2 depending on the temporary prefix.

This is not a breaking change per se but it's good to keep in mind.


src/common/paths.nim
```diff
@@ -99,10 +99,25 @@ let
     "CODETRACER_C_TRACE_OBJECT_FILE_PATH",
     linksPath / "lib" / "trace.o")
 
+when defined(ctmacos):
+  let codetracerTmpPath* = env.get("HOME") / "Library/Caches/com.codetracer.CodeTracer/codetracer"
+  let codetracerCache* = env.get("HOME") / "Library/Caches/com.codetracer.CodeTracer/codetracer_cache"
+else:
+  let tmpFolder = env.get("TMPDIR",
```
alehander92, line 106: is there a document that specifies this order

src/db-backend/src/paths.rs
```diff
@@ -0,0 +1,42 @@
...
+        {
+            PathBuf::from(env::var("TMPDIR").unwrap_or(
+                env::var("TEMPDIR").unwrap_or(
+                    env::var("TMP").unwrap_or(
+                        env::var("TEMP").unwrap_or(
+                            "/tmp".to_string()
+                        )
+                    )
+                )
+            )).join("codetracer/codetracer")
```
alehander92: why `codetracer/codetracer` twice?

src/db-backend/src/paths.rs
```diff
@@ -0,0 +1,42 @@
...
+        };
+        Self {
+            tmp_path: PathBuf::from(&tmpdir),
+            client_socket_path: PathBuf::from(&tmpdir).join("ct_client_socket"),
+            socket_path: PathBuf::from(&tmpdir).join("ct_socket"),
+        }
+    }   
+}
+
+pub static CODETRACER_PATHS: LazyLock<Mutex<Paths>> = LazyLock::new(|| {
```
alehander92, line 40: can we pass an object with the paths everywhere, so we don't need `lock/?` ? probably not, maybe that's the better option

alehander92 (commented):
good work, some clarification comments

src/common/paths.nim
Madman10K: TMP and TMPDIR are mainly used on Unix-like operating systems. Windows also has them but also uses TEMP and TEMPDIR. MSDN: <https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-gettemppatha>

alehander92: but does any of those have a priority? is there a reason we're trying them in this order (or is just our choice)

src/db-backend/src/paths.rs
Madman10K: There are too many calls to `Core::run_dir()`. To pass paths every time would be too annoying. Also, previously the paths were global variables too

src/db-backend/src/paths.rs
Madman10K: It's the same setup as in `common/paths.nim`. The idea is that macOS already forces us to put everything codetracer in a codetracer folder, instead of polluting `$TMP` with multiple codetracer-related files or folders. It's easier doing it this way instead of having special handling for Linux or any other platform. It should be cleaner for us.

The naming might be off though. Maybe the names of the inner folder need to be changed?

alehander92: hm, i see: it was a bit simpler for manual work to have a shorter path, otherwise it makes sense; maybe good just commands are sufficient; it's also possible to have `$TMP/ct/ct` or `$TMP/codetracer/ct` but maybe this would be confusing: explicit `$TMP/codetracer/` seems correct, and then the inner folder, not sure

src/common/paths.nim
Madman10K: TEMP and TEMPDIR are fallbacks that's why they are used last. No reason for the other 2

alehander92: here TEMP, TEMPDIR are used before `TMP`

alehander92: but i assume that's because it's unlikely to use them in the same OS with TMP

Madman10K:
Sasho kaza che e ok

PR #136, opened 2025-05-27
fix(ptr-values): make address string

Ro6afF:
The frontend expects that the value is string, but but previously number was passed, which caused exceptions


SersemPeca (approved):
LGTM

PR #147, opened 2025-05-29
Run receiver in a separate thread

alehander92:
## Summary
- start receiver loop in its own thread
- dispatch socket messages to a handler thread via a task channel
- adapt virtualization-layers binary to the new threaded receiver

## Testing
- `cargo build`
- `cargo test`
- `cargo clippy` *(fails: `clippy` component not installed)*


alehander92:
it seems correct.
but i think this will be outdated because of the dap changes: after them or as part of them we can write or generate code that separates the receiver there in this way if we need

PR #148, opened 2025-05-29
Implement Ruby each loop detection

alehander92:
## Summary
- detect Ruby `each` loops by looking at `call` nodes that have a block with method name `each`
- document Ruby loop insight

## Testing
- `cargo build`
- `cargo test`
- `cargo clippy` *(fails: `cargo-clippy` not installed)*


alehander92:
closing, superseded by #164

PR #158, opened 2025-05-30
Get commit hashes and repo names from flake.lock

### LINKED ISSUE #156: Sync hashes for dependencies between the non-nix-build and the nix build
During the debugging process of #128 the hashes for some of the dependencies of the non-nix-build being out of date caused a nasty bug. We need a way to keep them in sync or to fetch them from only 1 place.

Madman10K:
Fixes #156

### DIFF
--- a/non-nix-build/install_nargo.sh
+++ b/non-nix-build/install_nargo.sh
@@ -2,8 +2,6 @@
 
 set -e
 
-WANTED_NARGO_REVISION=8584e105549d6daa29d7b2be83bd5883016cfbd7
-
 if command -v nargo &> /dev/null; then
   echo nargo is already installed
   exit 0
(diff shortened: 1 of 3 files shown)


non-nix-build/install_nargo.sh
```diff
@@ -14,10 +12,19 @@ fi
...
+out=$(PWD=$(realpath ../../) ../find_git_hash_from_lockfile.py noir)
+commit=$(echo "${out}" | grep -v "github.com")
+repo=$(echo "${out}" | grep "github.com")
+folder="noir"
 
+mkdir ${folder} || echo "Folder already exists"
+cd "${folder}"
+if [ $(git rev-parse HEAD) != "${commit}" ]; then
+  cd ../
+  rm -rf "${folder}"
```
alehander92, line 24: if it's the same repo, we can fetch/checkout the commit and not replacing/cloning the repo again (i remember @zah did something like this in build scripts before)

if it's different, we can add a remote, but i agree sometimes it might be simpler to understand if it's replaced in this case (and if the repo is not that big)

non-nix-build/install_wazero.sh
```diff
@@ -2,15 +2,25 @@
...
+commit=$(echo "${out}" | grep -v "github.com")
+repo=$(echo "${out}" | grep "github.com")
+folder="codetracer-wasm-recorder"
+
+
+mkdir "${folder}" || echo "Folder already exists"
+cd "${folder}"
+if [ $(git rev-parse HEAD) != "${commit}" ]; then
+  cd ../
+  rm -rf "${folder}"
```
alehander92, line 19: same here

alehander92 (commented):
very good idea

my suggestions for more optimal git checkout cases are not critical, as they're just an optimization

non-nix-build/install_nargo.sh
Madman10K: So I just tried making an implementation using checkout and fetching. Unfortunately, even if using http, fetching requires a username and password to be set in git and for the user to type their credentials into the terminal.

This is obviously not fine, because:

1. People have to set up their git with a github account to build our app
1. Most people don't expect for there to be prompts in the middle of a build script

I guess we might have to keep the current workflow unfortunately :/

alehander92: yes, we don't want setup or a dialog, we might need to debug what's happening. it's ok like this for now probably  as this wasn't more optimal before IIRC

PR #163, opened 2025-05-30
Handle panics in the main thread of the db-backend

### LINKED ISSUE #153: DB-backend not reporting panics correctly from the main thread
While debugging #128 we found out that there was a panic in a certain part of the db-backend that was not being reported correctly. This was because there was no signal handler to report it properly on the main thread. Such a handler is already there for other threads that the db-backend spawns.

Madman10K:
Fixes #153.

### DIFF
--- a/src/db-backend/src/main.rs
+++ b/src/db-backend/src/main.rs
@@ -17,7 +17,7 @@ use std::sync::{mpsc, LazyLock, Mutex};
 use std::thread;
 use std::time::Instant;
 use std::{error::Error, panic};
-
+use std::panic::PanicHookInfo;
 use clap::Parser;
 use log::{error, info};
 use task::{EventId, EventKind, Notification, NotificationKind};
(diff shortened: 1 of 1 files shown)


src/db-backend/src/main.rs
```diff
@@ -66,7 +66,31 @@ struct Args {
 
 static TX2: LazyLock<Mutex<Option<mpsc::Sender<Response>>>> = LazyLock::new(|| Mutex::new(None)); // 🤮
 
+// Already panicking so the unwraps won't change anything
```
alehander92, line 69: it's good that they are used after the `error!` log!

src/db-backend/src/main.rs
```diff
@@ -66,7 +66,31 @@ struct Args {
...
+#[allow(clippy::unwrap_used)]
+fn panic_handler(info: &PanicHookInfo) {
```
alehander92, line 71: does this re-raise the panic?

alehander92 (commented):
nice!

alehander92 (approved):
well done!

src/db-backend/src/main.rs
alehander92: https://doc.rust-lang.org/std/panic/fn.set_hook.html it seems it's just a hook, which is what we need here, 
> Panics if called from a panicking thread.

PR #169, opened 2025-06-02
Use ruby from homebrew instead of the bundled macOS ruby 

### LINKED ISSUE #168: CodeTracer cannot record ruby on macOS
CodeTracer fails to record a ruby program. When launching the `/Applications/CodeTracer.app/Contents/MacOS/bin/ruby` binary manually to check if it works it kills itself with SIGKILL.

This is because we're not using the right ruby binary. It turns out that macOS already bundles a ruby framework, however this framework is Ruby v2.6(dated 2018) and is compiled in such a way that it cannot be used outside of its directory(`/usr/bin/`). One can create a symlink and this solves the issue but we get errors in our own code due to how ancient this version of Ruby is.

To fix the issue, the macOS scripts need to copy the ruby binary from homebrew. This binary is guaranteed to be modern enough that CodeTracer can use it without problems. The issue with SIGKILL cannot be reproduced with the modern homebrew binary as well.

A PR is on the way.

Madman10K:
Fixes #168

### DIFF
--- a/non-nix-build/build_in_simple_env.sh
+++ b/non-nix-build/build_in_simple_env.sh
@@ -38,7 +38,10 @@ bash build_db_backend.sh
 
 # setup/copy/link other files
 cp $ROOT_DIR/resources/electron $DIST_DIR/bin/
-cp $(which ruby) $DIST_DIR/bin/ruby
+
+# The built-in macOS ruby binary is too old and has to be hacked around
+cp $(brew --prefix ruby)/bin/ruby $DIST_DIR/bin/ruby
+
 cp $(which ctags) $DIST_DIR/bin/ctags
 cp $ROOT_DIR/libs/codetracer-ruby-recorder/src/*.rb $DIST_DIR/src/
 cp $ROOT_DIR/src/helpers.js $DIST_DIR/src/helpers.js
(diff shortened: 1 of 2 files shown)


non-nix-build/build_in_simple_env.sh
```diff
@@ -38,7 +38,10 @@ bash build_db_backend.sh
 
 # setup/copy/link other files
 cp $ROOT_DIR/resources/electron $DIST_DIR/bin/
-cp $(which ruby) $DIST_DIR/bin/ruby
+
+# The built-in macOS ruby binary is too old and has to be hacked around
+cp $(brew --prefix ruby)/bin/ruby $DIST_DIR/bin/ruby
```
alehander92, line 43: sounds good: i only find https://github.com/rbenv/rbenv/issues/343#issuecomment-14008198 as a possible problem: is it an issue, or would we maybe hope to think of a more stable solution longterm

Madman10K: This is no longer an issue. The symlink will not become stale because this is the output: `/opt/homebrew/opt/ruby`

alehander92: ok

SersemPeca (approved):
LGTM

PR #195, opened 2025-06-05
refactor(src/common_types): Part 1: split common_types.nim into multiple files

Madman10K:
This PR is part of a series of PRs that aim to refactor the way we deal with type sharing between the frontend and backends.

This PR aims to split the file into multiple smaller files. This should make development easier and should allow us to remove dead code, as well as moving code that is no longer common into its own domain.

The first commit of this PR aims to just split the `common_types.nim` file, without removing large portions of dead or dangling code.

Additional follow-up commits that aim to document types with ambiguous names are also planned.


src/common/common_types/tokens.nim
SersemPeca, line 1: Maybe move this to `language_features` ?

src/common/common_types/debugger_features/flow.nim
SersemPeca, line 1: This should go into something like `codetracer_features`

src/common/common_types/debugger_features/frontend.nim
SersemPeca, line 1: This should also go to into something like `codetracer_features`

src/common/common_types/debugger_features/shell.nim
SersemPeca, line 1: This should also go to into something like `codetracer_features`

alehander92 (commented):
looks good, some things look much better in an ordered way: not entirely sure if it breaks something in rr backend( I think Pavel can fix it independently, but let's still ping him before merging it); also not sure how tested for a regression it is

alehander92 (approved):
approving: on rr-backend we'll try to fix the breakages later

PR #198, opened 2025-06-05
Fix selectable footer status text

Franz-Fischbach:
Bug: The elements in the footer can be selected as shown in the screenshot below:
image

## Summary
- prevent selection of status bar items in the UI

## Testing
- `cargo test` in `src/db-backend`
- `cargo clippy` in `src/db-backend`
- `cargo build` in `src/db-backend`

------
https://chatgpt.com/codex/tasks/task_b_6841d0a52ccc832abf0aad675cd175c3


alehander92:
the fix seems very good!

however maybe we want to forbid accidental selection? if i want to explicitly copy and paste something there, maybe it should be possible? (but maybe not, i am wondering about that:it should be definitiely possible to select some longer parts of status bar like path, branch)

Issue #151, opened 2025-05-30
Can not scroll to the last value in scratchpad


Franz-Fischbach:
When there is a longer list of items in the scratchpad the last element can not always be seen as the scroll does not allow for more movement.

![Image](https://github.com/user-attachments/assets/3b1c2cf3-0a5d-4fb7-8536-ef78be244f4e)

Steps to reproduce:

1. Open CodeTracer using any program.
2. Start adding items to the scratchpad.
3. Change the size of the scratchpad panel

Expected result:
It should always be possible to scroll to the last item of the scratchpad.

Actual result:
Depending on the size of the container it is not always possible to scroll to the last record in the scratchpad. In some cases the last record is only partly visible.

Issue #146, opened 2025-05-29
Bind history forward / history backward to m4, m5 buttons


Franz-Fischbach:
When a new user was testing CodeTracer the following question was raised:
"Can we jump back and forth according to where I previously jumped?"
The answer to this question is yes, but it was not very intuitive.

One of the suggestions was to bind the history back and history forward buttons to the mouse m4 and m5 buttons similar to how a browser works.

![Image](https://github.com/user-attachments/assets/92bd25be-b526-43f1-8183-8f9089207cbc)

Issue #145, opened 2025-05-29
Can not use backspace on omniscient loop control iterations


Franz-Fischbach:
The inability to use backspace on the loop selection field is confusing for new users who want to type in another value to change the current iterration.

![Image](https://github.com/user-attachments/assets/91014c99-323a-4820-b7c4-a2c43a72aa8e)

Change the behavior and design so that so that the user can still see on which iteration CodeTracer is currently at but still be able to write freely into the field.

Issue #144, opened 2025-05-29
Tool tips for debug buttons appear too slowly


Franz-Fischbach:
The tool tips for the debug buttons appear after 2-3 seconds.
For new users who are new to CodeTracer this delay is rather irritating.

![Image](https://github.com/user-attachments/assets/298ecdc9-01f3-40d9-9fb5-10812b269eb7)

Use the time it takes google chrome to show a link on hover as an oracle for how long of a delay is acceptable.

![Image](https://github.com/user-attachments/assets/98ff02ad-5541-41fe-aff4-fedec9da82ea)

PR #209, opened 2025-06-09
fix: Frontend issues 01

pxor:
- Scratchpad height scroll #151 
- Change iterations for loops to be deletable instead of a placeholder value #145 
- Change the transition delay for the custom tooltips #144 
- Add support for browser back and browser forward mouse buttons #146
- Disable status bar text select


alehander92 (approved):
good work!

PR #218, opened 2025-06-10
feat(nix): Experimental dev shells for more minimal builds

zah:


nix/shells/rust-backend.nix
```diff
@@ -0,0 +1,63 @@
...
+        tree-sitter
+
+        # Useful for development
+        vim
+        delta
+      ];
+
+      shellHook = ''
+        echo "🦀 Rust backend development environment"
+        echo "Working directory: db-backend/"
```
alehander92, line 50: we have more rust code: also small-lang, tui, eventually runtime_tracing(but it's a submodule): i wonder if we should name this more generally `rust-shell` or something like this

scripts/update-tup-config
```diff
@@ -0,0 +1,44 @@
...
+
+# Script to update or add key-value pairs in tup.config files
+# Usage: update-tup-config [config_path] [key] [value]
+
+set -euo pipefail
+
+if [ $# -ne 3 ]; then
+    echo "Usage: $0 <config_path> <key> <value>"
+    echo "Example: $0 src/build-debug/tup.config WITHOUT_DB_BACKEND 1"
+    exit 1
```
alehander92, line 11: i get a `tup error: Non-comment line in tup config doesn't begin with "CONFIG_"
Line was: 'WITHOUT_DB_BACKEND=1'
` when i try it locally

nix/shells/rust-backend.nix
```diff
@@ -0,0 +1,63 @@
...
+in
+  with pkgs;
+    mkShell {
+      packages = [
+        # Rust toolchain
+        cargo
+        rustfmt
+        clippy
+        rust-analyzer
+
```
alehander92, line 16: we need `tup` here as well

nix/shells/rust-backend.nix
```diff
@@ -0,0 +1,63 @@
...
+        binutils
+        openssl
+        pcre
+        glib
+        libelf
+
+        # For inspecting and debugging
+        pstree
+        hexdump
+        tree-sitter
```
alehander92, line 41: currently we don't enable by default(they need `cargo test -- --ignored` iirc) some tests/similar that depend directly on e.g. ruby recorder, as it's not setup in codex: not sure if we should directly it here, or make the tests depend on `small-lang` or to include ruby/noir/ct in a separate `ct` devshell

alehander92: we can also commit traces, which we did in the `feat-dap` branch, but not sure if this is scalable/better

PR #227, opened 2025-06-12
Stylus subcommand

Ro6afF:
* Implement `ct arb deploy` and `ct arb replay` subcommands
* Use rustup as provider for rust toolchain in devshell
* Bump wazero version

### DIFF
--- a/src/ct/codetracerconf.nim
+++ b/src/ct/codetracerconf.nim
@@ -9,7 +9,8 @@ import
 type
   ArbCommand* {.pure.} = enum
     explorer,
-    replay
+    replay,
+    deploy
 
   StartupCommand* {.pure.} = enum
     noCommand,
@@ -227,7 +228,7 @@ type
         desc: "Arguments for record",
         longDesc: "longer description for record"
       .} : seq[string]
-    of replay:
+    of StartupCommand.replay:
      replayTraceId* {.
         name: "id",
         desc: "a trace id"
@@ -317,11 +318,13 @@ type
       .}: ArbCommand
       of explorer:
         discard
-      of replay:
+      of ArbCommand.replay:
         arbReplayTransaction* {.
           argument
           desc: "Hex-encoded transaction hash"
         .}: string
+      of deploy:
+        discard
     # of `import`:
     #   importTraceZipPath* {.
     #     argument
--- a/src/ct/codetracerconf.nim
+++ b/src/ct/codetracerconf.nim
@@ -9,6 +9,7 @@ import
 type
   ArbCommand* {.pure.} = enum
     explorer,
+    record,
     replay,
     deploy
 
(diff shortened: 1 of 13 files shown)


src/ct/codetracerconf.nim
```diff
@@ -300,6 +307,30 @@ type
...
       .}: bool
+    of arb:
+      arbitrumRpcUrl* {.
+        name: "arbitrum-rpc-url"
+        desc: "Arbitrum Node JSON-RPC URL"
+        defaultValue: "localhost"
+      .}: string
+      case arbCommand* {.
+        command,
+        defaultValue: explorer
```
zah: You can add a `noCommand` enum entry to the `ArbCommand` enum and make that the default.
This allows you to handle the case where the user launches CodeTracer with just `ct arb` (I think this could be an error for now. It may launch something in the future).

src/ct/launch/electron.nim
```diff
@@ -3,7 +3,17 @@ import std/[sequtils, os, osproc, strutils, strtabs ],
   ../globals,
   cleanup
 
-proc launchElectron*(args: seq[string] = @[], trace: Trace = nil, recordCore: bool = false, test: bool = false): bool =
+type
+  ElectronLaunchMode* {.pure.} = enum
+    None
```
zah: `Default` could be a better name for this -- yes, I know I chose `None` :)

src/ct/stylus/arb_node_utils.nim
```diff
@@ -0,0 +1,101 @@
+import std/[httpclient, json, strformat, strutils, sequtils, parseutils, times, sets, os]
+
+const DEFAULT_NODE_URL* = "http://localhost:8547"
+
+# TODO: get name from config? Maybe use SQLite?
+const CONTRACT_WASM_PATH* = getHomeDir() / ".local" / "share" / "codetracer" / "contract-debug-wasm"
+const EVM_TRACE_DIR_PATH* = getTempDir() / "codetracer"
+
+proc jsonRpcRequest(methodParam: string, params: JsonNode): JsonNode =
```
zah: Think about error checking in this module.
The code should either have `{.raises: .}` annotations to make it clear how it can fail or it should handle all possible exceptions and translate them to `Result` return value.

src/ct/stylus/deploy.nim
```diff
@@ -0,0 +1,107 @@
+import std/[json, os, osproc, streams, strutils]
+import arb_node_utils
+
+proc doDeploy(): string =
```
zah: Ditto (not sufficient error checking).

src/ct/stylus/record.nim
```diff
@@ -0,0 +1,58 @@
+import std/[httpclient, json, os, osproc, times, streams]
+import arb_node_utils
+import ../../common/[paths, trace_index, types]
+import ../trace/record
+
+proc getEvmTrace(hash: string): string =
```
zah: Ditto (not sufficient error checking).

PR #229, opened 2025-06-16
fix: Some more frontend issues and bugs

pxor:


alehander92:
ok, fixing in rr-backend separately

PR #234, opened 2025-06-23
feat: binary trace format

nickysn:
This adds support for the binary trace format, introduced in runtime_tracing version 0.12.0. It is already supported by the Noir trace generator in:

https://github.com/noir-lang/noir/pull/7758

The previous JSON format is still supported for backwards compatibility, and to support the language backends that still haven't implemented the binary format.


alehander92 (commented):
looks great, we just need to figure out how to setup `capnp` binary for the various builds (the only weird behavior  is for `tup` in devshell currently)

nickysn:
I have read the CLA Document and I hereby sign the CLA

PR #235, opened 2025-06-25
Handle load-locals in DAP server

pxor:
## Summary
- add `load-locals` request support in the DAP server
- forward variables obtained via `handler.load_locals`
- update DAP tests to account for optional launch fields

## Testing
- `cargo clippy`
- `cargo build`


alehander92:
wait: we need to namespace it like `ct-load-locals` or `codetracer-load-locals` or `ct/load-locals`

PR #239, opened 2025-07-02
feat: Add long omniscience values to be truncated and a button to expand them

pxor:
The event log is updated to respect the new line on the event output
and we add a show/hide omniscience value toggle button

### DIFF
--- a/src/db-backend/src/event_db.rs
+++ b/src/db-backend/src/event_db.rs
@@ -2,7 +2,7 @@ use std::cmp::min;
 use std::collections::HashMap;
 use std::error::Error;
 
-use log::{error, info};
+use log::error;
 use runtime_tracing::{EventLogKind, StepId, TypeKind};
 
 use crate::task::{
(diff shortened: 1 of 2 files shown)


src/frontend/ui/flow.nim
```diff
@@ -1270,6 +1270,23 @@ proc flowSimpleValue*(
       style=style
     )
   ):
+    span(
+      class = &"flow-{flowMode}-value-name flow-view-more-button flow-hide-content",
+      style = style,
+      onmousedown = proc(e: Event, v: VNode) =
+        let targetId = &"flow-{flowMode}-value-box-{i}-{stepCount}-{name}"
+        let target = document.getElementById(targetId)
+        if target != nil:
```
alehander92: isNil

src/frontend/ui/flow.nim
```diff
@@ -1270,6 +1270,23 @@ proc flowSimpleValue*(
...
+          kout e
```
alehander92: comment out before merging or convert to a `clog`

Ro6afF (approved):

PR #251, opened 2025-07-14
docs: Refactor intro and installation sections

tzanko-matev:


docs/book/src/installation.md
```diff
@@ -0,0 +1,65 @@
...
+
+### Linux
+You can download the Linux AppImage from our website:
+[![Download AppImage](https://img.shields.io/badge/Download-Linux%20AppImage-blue?style=for-the-badge)](https://downloads.codetracer.com/CodeTracer-25.05.1-amd64.AppImage)
+
+### macOS
+macOS binaries are temporarily unavailable but will be restored soon.
+
+> [!TIP]
+> You can place the downloaded app in a location of your choosing (e.g., the `Applications` folder on macOS or `~/.local/share/applications` on Linux).
```
Madman10K: This is written as part of the macOS part so it should only have instructions for macOS.

Otherwise it would be better if the path for Linux was `~/.local/bin`, since `~/.local/share/applications` is not normally in the user's path.

The `~/.local/share/applications` directory is used for the desktop file, which users are not required to install manually, since the user can decided to install it when clicking on the `Install` button in the first-time startup prompt.

docs/book/src/installation.md
```diff
@@ -0,0 +1,65 @@
...
+When you launch CodeTracer for the first time, it will prompt you to complete the remaining installation steps, such as adding the command-line utilities to your PATH.
+
+> [!CAUTION]  
+> Upon the first launch, macOS users will see the error message "CodeTracer is damaged and can't be opened". To resolve this problem, please execute the command `xattr -c <path/to/CodeTracer.app>`. We expect this inconvenience will be remedied soon through our enrollment in the Apple Developer program that will ensure CodeTracer is properly signed and whitelisted by Apple. See https://discussions.apple.com/thread/253714860?sortBy=rank for more details.
+
+## Installation from Source
```
Madman10K: Source should be in lower case :sweat_smile:

docs/book/src/installation.md
```diff
@@ -0,0 +1,65 @@
...
+
+### Prerequisites
+
+On systems that are not NixOS, you need to install `direnv` and `nix`.
```
Madman10K, line 33: This is no longer entirely correct. These steps were true back when we wrote this document in January when we lacked a macOS build.

The situation with macOS is not satisfactory but the current arrangement is:

1. The current instructions work for Linux, so no need to change them
1. On macOS building is done by running `./non-nix-build/build.sh` from the root of the cloned repository.

Git, bash and homebrew are the only dependencies of the macOS build.

On macOS the resulting binary is installed in `./non-nix-build/CodeTracer.app/Contents/MacOS/bin/ct`, while the DMG installer is `./non-nix-build/CodeTracer.dmg`.

Madman10K (changes requested):
Overall a good PR :+1:

docs/book/src/installation.md
tzanko-matev: Please check now

docs/book/src/installation.md
tzanko-matev: Done

docs/book/src/installation.md
tzanko-matev: I added the mac instructions. Please check now

PR #258, opened 2025-07-16
Fix AppImage on non-NixOS systems that have Nix installed

### LINKED ISSUE #97: AppImage has uses incompatible libc from nix store when nix is installed on a non-NixOS system
I set up a new Arch Linux machine today. I downloaded Nix for local development, but I also wanted to do testing on the new PKGBUILD I'm preparing that uses the AppImage under the hood.

However, when launching the AppImage, the AppImage tries to load a libc library from the nix store, instead of from my `/usr/lib` directory. This leads to a crash due to some incompatibility between the 2 libc versions:
```
ct: symbol lookup error: /nix/store/nqb2ns2d1lahnd5ncwmn6k84qfd7vx2k-glibc-2.40-36/lib/libc.so.6: undefined symbol: __nptl_change_stack_perm, version GLIBC_PRIVATE
```
For reference, my system libc version is 2.41.

I suspect that this is either because nix forces itself to be the first dynamic library path to be loaded, or we have some custom functionality in the AppImage build process that prefers libraries from the nix store

> [!NOTE]
> The only application I have that uses nix is the local build of codetracer, so this might also be a useful lead

> [!NOTE]
> The AppImage version is 25.03.1 because that's the only one we have packaged as of yet

Madman10K:
Fixes #97 

After playing around inspecting the produced AppImage binary I found that we were not setting the rpath correctly, so I added some additional entries where we should be looking.

> [!CAUTION]
> The patch only works for the latest master so retroactively applying it to the current latest release is not possible.

### DIFF
--- a/appimage-scripts/build_appimage.sh
+++ b/appimage-scripts/build_appimage.sh
@@ -225,6 +225,7 @@ fi
 ARCH=$APPIMAGE_ARCH appimagetool "${APP_DIR}" CodeTracer.AppImage
 
 patchelf --set-interpreter "${INTERPRETER_PATH}" "${ROOT_PATH}"/CodeTracer.AppImage
+patchelf --set-rpath "/lib:/lib64:/usr/lib:/usr/lib64:$(patchelf --print-rpath "${ROOT_PATH}"/CodeTracer.AppImage)" "${ROOT_PATH}"/CodeTracer.AppImage
 
 echo "============================"
 echo "AppImage successfully built!"


Madman10K:
This is not an issue, because this is the RPATH for the AppImage binary:
```
/nix/store/cl30k9b1i6166yw7gj06fzn2fvfxhaka-appimagekit-unstable-2020-12-31/lib:/nix/store/l5g2v1jgfyf3j0jp9iv5b79fi8yrwzpp-zlib-1.3.1/lib:/nix/store/bmzls7bipx954xsnbgshh8kbi488zy5m-xz-5.6.3/lib:/nix/store/nqb2ns2d1lahnd5ncwmn6k84qfd7vx2k-glibc-2.40-36/lib:/nix/store/4gk773fqcsv4fh2rfkhs9bgfih86fdq8-gcc-13.3.0-lib/lib
```

or for more readability:

1. `/nix/store/cl30k9b1i6166yw7gj06fzn2fvfxhaka-appimagekit-unstable-2020-12-31/lib`
1. `/nix/store/l5g2v1jgfyf3j0jp9iv5b79fi8yrwzpp-zlib-1.3.1/lib`
1. `/nix/store/bmzls7bipx954xsnbgshh8kbi488zy5m-xz-5.6.3/lib`
1. `/nix/store/nqb2ns2d1lahnd5ncwmn6k84qfd7vx2k-glibc-2.40-36/lib`
1. `/nix/store/4gk773fqcsv4fh2rfkhs9bgfih86fdq8-gcc-13.3.0-lib/lib`

As you can see these paths are the paths for the dependencies that the AppImage format needs to start. If the AppImage was not built under NixOS, the path would probably be `/usr/lib:/usr/lib64` or something similar. You can prove this by looking at nix packages of applications that are distributed as appimages:
```nix
unpackPhase = ''
    cp $src appimagetool-x86_64.AppImage
    chmod u+wx appimagetool-x86_64.AppImage
    patchelf \
      --set-interpreter "$(cat $NIX_CC/nix-support/dynamic-linker)" \
      --set-rpath ${fuse}/lib:${zlib}/lib \
      appimagetool-x86_64.AppImage

    ./appimagetool-x86_64.AppImage --appimage-extract
  ''
```

Meanwhile, the codetracer binary has its own RPATH that is not controlled by this RPATH. Subsequently, codetracer looks in the correct directories by preferring the embedded nix store.

Madman10K:
Already fixed in main

PR #266, opened 2025-07-18
Try to fix ct hanging processes problem

alehander92:
try to fix process hanging for ct_wrapper/started from welcome screen case
  Dimo (@Ro6afF ) and I mostly in a shared session produced debugging/solving this and this code
  
   also: tuprules change here: Dimo made most of the nim targets to build with debuginfo for easier debugging for the tup(for now always dev) mode

### DIFF
--- a/src/ct/launch/launch.nim
+++ b/src/ct/launch/launch.nim
@@ -147,7 +147,6 @@ proc runInitial*(conf: CodetracerConf) =
       of ArbCommand.deploy:
         deployStylus()
       of ArbCommand.listRecentTx:
-        # discard getTrackableTransactions()
         let transactions = getTrackableTransactions()
         let res = Json.encode(transactions)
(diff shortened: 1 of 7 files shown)


src/ct/launch/launch.nim
```diff
@@ -147,7 +147,9 @@ proc runInitial*(conf: CodetracerConf) =
       of ArbCommand.deploy:
         deployStylus()
       of ArbCommand.listRecentTx:
-        let res = Json.encode(getTrackableTransactions())
+        # discard getTrackableTransactions()
```
Madman10K: Comment not needed

alehander92: sorry, leftover

alehander92: ok, fixed

PR #274, opened 2025-07-21
fix(documentation): rearange user manual book

Franz-Fischbach:
Rearanged readme book section


docs/book/src/usage_guide/getting_started/stylus_and_wasm.md
```diff
@@ -0,0 +1,68 @@
...
+   run-nitro-devnode
+   ```
+
+2. **Deploy the demo contract**
+
+   Open a new terminal, navigate to the directory of your program and let CodeTracer deploy the smart contract.
+
+   ```bash
+   cd ui-tests/programs/stylus_fund_tracker
+   ct arb deploy
```
Ro6afF: Maybe we should add `nix develop` or `direnv allow`  and `just build-once` before this, so that we are sure, that there is the latest version of ct in the path?

docs/book/src/usage_guide/getting_started/stylus_and_wasm.md
```diff
@@ -0,0 +1,68 @@
+## Stylus / WASM
```
Ro6afF: I think, that it would be better to have separate entries for Stylus and WASM programs. That is because the workflow for Stylus is what is described in this document, but the workflow for WASM programs is closer to the Ruby workflow.

Ro6afF (changes requested):

docs/book/src/usage_guide/getting_started/stylus_and_wasm.md
alehander92: for WASM binaries*: high level language programs that are compiled to WASM (because we can have possibly in the future support for WASM bytecode directly as well, which is more like a WASM program, sorry for the nitpicking)

tzanko-matev (approved):

PR #278, opened 2025-07-21
Improve the release workflow

Madman10K:
This PR does the following:

1. Fixes a number of bugs with our conditions
1. Now pushes a `latest` version of the codetracer appimage and dmg when a release is ready, circumventing the need for manual version changes
1. The dmg and appimage are signed with our organisation's pgp key

### DIFF
--- a/.github/workflows/codetracer.yml
+++ b/.github/workflows/codetracer.yml
@@ -97,7 +97,27 @@ jobs:
           substituters: ${{ vars.SUBSTITUTERS }}
 
       - run: "nix develop .#devShells.x86_64-linux.default -c ./ci/lint/ui-tests.sh"
-
+  push-gpg-public-key:
+    runs-on: [self-hosted, nixos]
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: "Import GPG key for signing commits"
+        id: import-gpg
+        uses: crazy-max/ghaction-import-gpg@v6
+        with:
+          gpg_private_key: ${{ secrets.CODETRACER_AUR_GPG_PRIVATE_KEY }}
+          passphrase: ${{ secrets.CODETRACER_AUR_GPG_PRIVATE_KEY_PASS }}
+          git_config_global: true
+          git_user_signingkey: true
+          git_commit_gpgsign: true
+      - name: "Upload public key"
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.R2_CODETRACER_BUCKET_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.R2_CODETRACER_BUCKET_ACCESS_KEY }}
+        run: |
+          nix develop .#devShells.x86_64-linux.default --command gpg --armor --export > CodeTracer.pub.asc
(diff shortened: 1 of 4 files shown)


.github/workflows/codetracer.yml
```diff
@@ -390,8 +390,8 @@ jobs:
 
           We're actively working on multiple exciting features, which are not yet fully released. Stay tuned!
 
-          [![Download AppImage](https://img.shields.io/badge/Download-Linux%20AppImage-blue?style=for-the-badge)](https://downloads.codetracer.com/CodeTracer-${{ github.ref_name }}-amd64.AppImage)
-          [![Download macOS](https://img.shields.io/badge/Download-macOS-blue?style=for-the-badge)](https://downloads.codetracer.com/CodeTracer-${{ github.ref_name }}-arm64.dmg)" >> release_changelog.md
+          [![Download AppImage](https://img.shields.io/badge/Download-Linux%20AppImage-blue?style=for-the-badge)](https://downloads.codetracer.com/CodeTracer-latest-amd64.AppImage)
```
alehander92: i am not sure about this: `CodeTracer-25.1-amd64.AppImage` is very different from `CodeTracer-27.1..` : here both will be `-latest`. 
 I think it's good that the artifacts have version numbers in their names, and the install phase can omit them (?), but maybe that's a wrong view: do people usually not put the version number in such github releases?

.github/workflows/codetracer.yml
```diff
@@ -172,7 +172,7 @@ jobs:
         run: "nix develop .#devShells.x86_64-linux.default --command ./ci/build/appimage.sh"
 
       - name: Upload artifact
-        if: startsWith(github.ref, 'refs/tags/') && ${{ !github.event.codetracer-ci }}
+        if: 'startsWith(github.ref, "refs/tags/") && ${{ !github.event.codetracer-ci }}'
```
alehander92: does this help with parsing the whole expression as a single yaml string? nice

alehander92:
> for manual version changes

we do need to manually update `src/ct/version.nim` for now though!

README.md
```diff
@@ -1,9 +1,8 @@
...
 [![CI Status](https://github.com/metacraft-labs/codetracer/actions/workflows/codetracer.yml/badge.svg?branch=main)](https://github.com/metacraft-labs/codetracer/actions/workflows/codetracer.yml)
 [![Discord](https://img.shields.io/discord/1326949714679038014?label=Discord&logo=discord&style=flat)](https://discord.gg/aH5WTMnKHT)
 
-[![Download AppImage](https://img.shields.io/badge/Download-Linux%20AppImage-blue?style=for-the-badge)](https://downloads.codetracer.com/CodeTracer-25.05.1-amd64.AppImage)
-<!-- FIXME! Temporary disabled due to broken image [![Download macOS](https://img.shields.io/badge/Download-macOS-blue?style=for-the-badge)](https://downloads.codetracer.com/CodeTracer-25.05.1-arm64.dmg) -->
+[![Download AppImage](https://img.shields.io/badge/Download-Linux%20AppImage-blue?style=for-the-badge)](https://downloads.codetracer.com/CodeTracer-latest-amd64.AppImage)
+[![Download macOS](https://img.shields.io/badge/Download-macOS-blue?style=for-the-badge)](https://downloads.codetracer.com/CodeTracer-latest-arm64.dmg)
```
alehander92, line 7: maybe we can duplicate or somehow point dynamically to the last image, so we can have both the version number in the filename, and a stable link for the latest one?

alehander92: ok, i think this is how it works

.github/workflows/codetracer.yml
alehander92: i think we have both now for releases

PR #293, opened 2025-07-23
Fix certain tooltips and popups being clipped in the first lines of code

### LINKED ISSUE #132: Variable information is clipped in the first lines of code
When a variable is defined in the first two lines of code, on hover the tool tip gets clipped as shown in the screenshot below.

Image

Madman10K:
Fixes #132


alehander92:
ok, well done

PR #300, opened 2025-07-25
ci/deploy/docs.sh: Fix GPG signing issue during CI

tzanko-matev:
The CI requires that commits are GPG signed, which breaks the deploy/docs.sh script. We've added an argument which should fix this.

Here's a script which reproduces the issue locally:

```sh
#!/usr/bin/env bash

# This script attempts to reproduce a GPG signing error seen in the CI.
# It should be run from the root of the codetracer repository.

set -e # Exit immediately if a command exits with a non-zero status.

echo "--- Starting GPG error reproduction script ---"

# Check if we are in the root of the repo
if [ ! -f ".git/config" ]; then
    echo "ERROR: This script must be run from the root of the repository."
    exit 1
fi

# The original script operates inside docs/book
if [ ! -d "docs/book" ]; then
    echo "ERROR: The 'docs/book' directory does not exist."
    exit 1
fi

pushd docs/book/ > /dev/null

# Store original git config values to restore them later
original_gpgsign=$(git config --get commit.gpgsign || true)
original_signingkey=$(git config --get user.signingkey || true)

# Cleanup function to restore git config and clean up worktree
cleanup() {
  echo "--- Cleaning up ---"

  # Go back to the worktree root if we are in the gh-pages dir
  if [[ "$(basename "$PWD")" == "gh-pages" ]]; then
    cd ..
  fi

  # Remove the worktree
  if [ -d "gh-pages" ]; then
      echo "Removing 'gh-pages' worktree..."
      git worktree remove --force gh-pages
  fi

  # Remove the local branch
  if git show-ref --verify --quiet refs/heads/gh-pages; then
      echo "Deleting 'gh-pages' branch..."
      git branch -D gh-pages
  fi

  # Restore original git config
  echo "Restoring original git config..."
  if [ -n "$original_gpgsign" ]; then
    git config commit.gpgsign "$original_gpgsign"
  else
    git config --unset commit.gpgsign
  fi

  if [ -n "$original_signingkey" ]; then
    git config user.signingkey "$original_signingkey"
  else
    git config --unset user.signingkey
  fi

  popd > /dev/null
  echo "--- Cleanup complete ---"
}

# Register the cleanup function to run on exit, error, or interrupt
trap cleanup EXIT ERR INT

echo "Setting up git config to reproduce the signing error..."
# Force GPG signing for commits
git config commit.gpgsign true
# Use the specific key from the CI error log
git config user.signingkey "7EAF585FB9B59164"

# The CI script does this, so we do it too for consistency.
git config user.name "Deploy from CI"
git config user.email "" # The CI script uses an empty email

# Clean up from any previous failed runs
echo "Pruning and removing any old 'gh-pages' worktree..."
git worktree prune
if [ -d "gh-pages" ]; then
    git worktree remove --force gh-pages
fi
if git show-ref --verify --quiet refs/heads/gh-pages; then
    git branch -D gh-pages
fi

# The CI script builds the book first. We'll just create a dummy file
# to ensure the `cp` command has something to copy.
mkdir -p book
touch book/index.html
echo "Dummy book content" > book/index.html
echo "Creating new orphan 'gh-pages' worktree..."
git worktree add --orphan -B gh-pages gh-pages

echo "Copying 'book' contents to 'gh-pages' worktree..."
cp -a book/. gh-pages/

cd gh-pages

echo "Staging files in 'gh-pages'..."
git add -A

echo "Attempting to create a commit..."
echo "This command is expected to fail with a GPG signing error."
echo "----------------------------------------------------------"

# This is the command that should fail
git commit -m 'deploy new book' # When we add --no-gpg-sign it works!!!

# If the script reaches here, the error was not reproduced.
echo "----------------------------------------------------------"
echo "UNEXPECTED: The commit was successful and the error was not reproduced."
echo "Please check your local GPG and git configuration."

# The trap will handle cleanup.
exit 1

```

### DIFF
--- a/ci/deploy/docs.sh
+++ b/ci/deploy/docs.sh
@@ -36,7 +36,7 @@ git config user.name "Deploy from CI"
 git config user.email ""
 cd gh-pages
 git add -A
-git commit -m 'deploy new book'
+git commit -m 'deploy new book' --no-gpg-sign
 git push origin +gh-pages
 cd ..
 


alehander92 (approved):
great reproduction again

makes sense; however is adding `-no-gpg-sign` a workaround? this way they stay unsigned, which is probably ok, just wasn't sure if there is any alternative approach

imo that's ok for merge

PR #303, opened 2025-07-25
Fix broken libzip and ruby location in macOS bundle

Madman10K:
This PR does the following:

1. Fixes usage of absolute path for the ruby binary
1. Informs users that they need to install ruby from homebrew
1. Fixes libzip being added to the wrong directory
1. Improves the reabability of the README tips and warnings
1. Applies the readability improvements from the README and adds them to the documentation with additional warnings about the ruby support


Madman10K:
Dynamic linking with libzip seems to be impossible unless we compile libzip from source so instead I enabled static linking to save us the effort. It also simplifies our build which is always a bonus

alehander92 (approved):

PR #305, opened 2025-07-28
feat: binary trace format "version 1", based on CBOR and Zstd seekable compression

nickysn:
This brings support for the new "version 1" of the binary trace format, which is based on CBOR + Zstd compression. All the previous formats (binary "version 0", based on capnproto, and the JSON format) are still supported and should continue to work seamlessly. In the case of binary, the exact version of the format is detected from the file header.


SersemPeca (approved):
LGTM.

Issue #297, opened 2025-07-24
Add a multiplexer : `backend_manager` that lets us manage/communicate from 1 client to N backend instances


alehander92:
discussed with @Ro6afF in a call

PR #307, opened 2025-07-29
feat: implement backend-manager

Ro6afF:
This PR implements #297

### DIFF
--- a/src/backend-manager/src/backend_manager.rs
+++ b/src/backend-manager/src/backend_manager.rs
@@ -40,13 +40,11 @@ impl BackendManager {
         let res1 = res.clone();
         let res2 = res.clone();
 
-        let mut socket_path = env::temp_dir(); // TODO: discuss what is the best place for the socket. Maybe /run?
-        socket_path.push("codetracer");
-        socket_path.push("backend-manager");
+        let socket_dir = env::temp_dir().join("codetracer").join("backend-manager");
 
-        create_dir_all(&socket_path).await?;
+        create_dir_all(&socket_dir).await?;
 
-        socket_path.push(std::process::id().to_string() + ".sock");
+        let socket_path = socket_dir.join(std::process::id().to_string() + ".sock");
         _ = remove_file(&socket_path).await;
 
         let mut socket_read;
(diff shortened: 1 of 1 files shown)


src/backend-manager/src/backend_manager.rs
```diff
@@ -265,192 +270,92 @@ impl BackendManager {
...
-                            return self.message_selected(message).await;
-                        }
-
-                        // SAFETY: The if above ensures that replay-id exists
-                        let replay_id = unsafe { args.get("replay-id").unwrap_unchecked() };
-
-                        if !replay_id.is_u64() {
-                            // Expected integer ID. IDK what this is. Forwarding...
-                            return self.message_selected(message).await;
+                        if let Some(Value::Object(obj_args)) = args {
```
alehander92, line 343: that's much better: still maybe we can directly map it but in any case an improvement

src/backend-manager/src/backend_manager.rs
```diff
@@ -0,0 +1,364 @@
...
+                        }
+                    }
+                }
+            }
+        });
+
+        tokio::spawn(async move {
+            let mut parser = DapParser::new();
+
+            let mut buff = vec![0; 8 * 1024];
```
alehander92, line 89: is this enough?

src/backend-manager/src/backend_manager.rs
```diff
@@ -0,0 +1,364 @@
...
+            None => Err(Box::new(InvalidID(id))),
+        }
+    }
+
+    pub async fn start_replay(
+        &mut self,
+        cmd: &str,
+        args: &[&str],
+    ) -> Result<usize, Box<dyn Error>> {
+        let mut socket_path = env::temp_dir(); // TODO: discuss what is the best place for the socket. Maybe /run?
```
alehander92: using something like path join , not push?

alehander92: (the current path itself sounds good!)

src/backend-manager/src/backend_manager.rs
```diff
@@ -0,0 +1,364 @@
...
+
+        let child = cmd.spawn();
+        let child = match child {
+            Ok(c) => c,
+            Err(err) => {
+                error!("Can't start replay: {err}");
+                return Err(Box::new(err));
+            }
+        };
+        sleep(Duration::from_millis(10)).await;
```
alehander92: flakey: can we have a better way to wait here?

src/backend-manager/src/backend_manager.rs
```diff
@@ -0,0 +1,364 @@
...
+                        error!("Can't send message to replay socket! Error: {err}");
+                    }
+                }
+            }
+        });
+
+        tokio::spawn(async move {
+            let mut parser = DapParser::new();
+
+            let mut buff = vec![0; 8 * 1024];
```
alehander92, line 215: ditto: what if we receive a bigger chunk in `read`

src/frontend/tests/testing_framework/test_helpers.nim
```diff
@@ -123,6 +123,7 @@ proc startApp*(programName: string, lang: Lang) {.async.} =
     echo fmt"error: no trace for {programName} and {lang} (for program {path})"
     quit(1)
   let traceId = trace.id
+  # TODO: discuss if this code is still relevant and how to rework it?
```
alehander92, line 126: good question, not sure if we use this indeed: you can ignore it for this PR

src/backend-manager/src/backend_manager.rs
Ro6afF: Moved the socket listener from db-backend to backend-manager

src/backend-manager/src/backend_manager.rs
Ro6afF: Yes, if there is more data, then it will be read in multiple chunks

PR #312, opened 2025-08-08
Feat: Integrate more action for the electron client with the new protocol

pxor:
This MR works on adding more integration with the new protocol:
- Support breakpoints adding/deleting
- Support tracepoints


alehander92:
looked at it while cooperating on it: ok

alehander92 (approved):

PR #321, opened 2025-08-12
Try to fix stepping from shortcuts

### LINKED ISSUE #319: Keyboard shortcuts not working
On the branch feat-extension-support the keyboard shortcuts like `F10`, `ctrl + p` and others do not work.

Steps to reproduce:

1. checkot feat-extension-support
2. Open CodeTracer using `ct run examples/noir_space_ship/`
3. Try using F10 to move to the next step
4. Try using ctrl + p to open the command palette

Expected result:
CodeTracer should react to the keyboard shortcuts

Actual result:
Nothing happens

alehander92:
eventually this can be made in a more clean way maybe:
  decide where this logic should stay

also we can make a custom `ct/step` or `ct/custom-step` request for max compatibility with the old step, but not sure if relevant for now

i think when ready, this resolves #319 , but have to check if there are problems with other shortcuts: some seemed to work, but there were some devtools errors : haven't checked if they were related. (and not checked many shortcuts yet)

### DIFF
--- a/src/frontend/renderer.nim
+++ b/src/frontend/renderer.nim
@@ -100,7 +100,7 @@ proc gotoLine*(line: int, highlight: bool = false, change: bool = false) {.expor
 proc lowAsm*(data: Data): bool
 proc highlightLine*(path: cstring, line: int)
 proc saveFiles*(data: Data, path: cstring = j"", saveAs: bool = false)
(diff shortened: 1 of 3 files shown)


alehander92:
with @pxor : fixed more stuff, should be ok now

Issue #324, opened 2025-08-12
Moving Golden Layout Groups Is Slow


Franz-Fischbach:
When Moving the golden layout delimiters, on every other move there is a 2-3 second delay.
 The first move is fast, the second slow, the third move is fast the fourth is slow and so on. 

Steps to reproduce:
1. Checkout feat-extension-support 
Note: as of writing, the bug does not reproduce on the main branch

2. Open CodeTracer using `ct run examples/noir_space_ship/ `
3. See screenshot

<img width="3200" height="1328" alt="Image" src="https://github.com/user-attachments/assets/f3300ed9-87c4-4ccc-ae3f-03a868a38efe" />

4. See screenshot

<img width="3200" height="1328" alt="Image" src="https://github.com/user-attachments/assets/c937a74c-0dd3-4165-9b47-ef258aea1ffa" />

5. See screenshot

<img width="3200" height="1328" alt="Image" src="https://github.com/user-attachments/assets/7bf7be18-9313-43f1-ab72-49e58482c4e7" />

6. See screenshot

<img width="3200" height="1328" alt="Image" src="https://github.com/user-attachments/assets/b9cde5e5-a398-473d-a5f8-2257824c20c2" />

Expected result:

Moving the panel delimiters should always be fast

Actual result:
Every other move has a 2-4 second delay

Franz-Fischbach:
Can not reproduce after fix

Issue #323, opened 2025-08-12
Removed items from scratchpad remain visible until change of focus


Franz-Fischbach:
When removing items from the scratchpad via the X button, the items stay visible until the focus is changed to another component.

Steps to reproduce:
1. Checkout feat-extension-support 
Note: as of writing, the bug does not reproduce on the main branch

2. Open CodeTracer using `ct run examples/noir_space_ship/ `
3. Click on one of the `calculate_remaining_shield_pct` funtion calls in the calltrace
4. Rightclick on the return value of the function and selct Add all values to the scratchpad as shown in the screenshot below

<img width="3200" height="1328" alt="Image" src="https://github.com/user-attachments/assets/126326bd-56cd-4978-93ed-4649f76d169b" />

5. There should be a new record in the scratchpad, press the X button

Expected result:

The record should dissapear immediately after pressing the X button

Actual result:

The scratchpad record stays there until the focus is moved to another component

alehander92:
(possibly fixed by https://github.com/metacraft-labs/codetracer/issues/323 : has to be checked)

Issue #322, opened 2025-08-12
Calltrace nesting visualisation offset is moved when interacting with the golden layout delimiters


Franz-Fischbach:
When resizing golden layout groups the calltrace nesting visualisation gets moved more than it should as shown in the screenshots below:

Steps to reproduce:
1. Checkout feat-extension-support 
Note: as of writing, the bug does not reproduce on the main branch

2. Open CodeTracer using `ct run examples/noir_space_ship/ `

3. See screenshot below

<img width="3200" height="1328" alt="Image" src="https://github.com/user-attachments/assets/522fec83-606a-40e4-bd16-bf68b24853cd" />

Expected result:

The calltrace nesting visualisation should move relative to the text inside the pane

Actual result:

See screenshots

<img width="3200" height="1328" alt="Image" src="https://github.com/user-attachments/assets/e046e6f4-cb6b-4058-82ef-b09e85ce427c" />

<img width="3200" height="1328" alt="Image" src="https://github.com/user-attachments/assets/9560f4e8-bc03-4459-9552-58d9b56ae96c" />

PR #325, opened 2025-08-13
Fix: Regressions for electron client after DAP integration

pxor:
- Fix minor issue with building for extension
- Fix mutation observer for calltrace #322 
- Fix scratchpad redraw issues #323
- Double component registration issue which might have been the issue with the slow golden layout movements #324 (more testing needs to be done on the matter)

### DIFF
--- a/src/common/ct_event.nim
+++ b/src/common/ct_event.nim
@@ -63,7 +63,7 @@ type
     CtSetupTraceSession,
     InternalLastCompleteMove,
     InternalAddToScratchpad,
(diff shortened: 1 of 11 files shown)


alehander92:
we are leaving a regression: now add to scratchpad won't automatically reopen the scratchpad panel if it is closed, this requires us to move this reopening logic to a middleware handler, but i want to leave this eventually for the future, maybe after this release, as it seems as a minor feature to me

alehander92 (approved):

PR #326, opened 2025-08-13
Fix: Current operation status

pxor:


alehander92 (approved):
there might be more emit-s/operations left to update the status for, but we can take a look at them in a future moment: i hope at least most cases that were working in the `main` are handled.(however some of them with less detail)

PR #327, opened 2025-08-13
Notification handling

pxor:
Working on fixing the notifications for the electron client. Currently reaching a compilation error:
```
Error: internal error: genTypeInfo(tyNil)
Traceback from system (most recent call last)
```


alehander92 (approved):
maybe some cases are left, but at least end already reached/similar and install notifications seems to work

Issue #341, opened 2025-08-19
define/autogenerate dap/ct extension types with a single source of truth


alehander92:
@lispegistus / @Ro6afF have proposed something like this in the past:
for now trying spent a bit of time: several days, where i spent several hours each, to use JSON schema for that ( or rust, or if this is hard: typescript; types autogenerating JSON schema) as a single source of truth;

generating rust(if needed) and nim (and eventually python) types from them with a custom script: `schema.js` 

i've tried https://quicktype.io/schema, 

it looks very well done, but I had some problems with the DAP schema ( https://github.com/glideapps/quicktype/issues/1137 it seems for some people it could work) and we need to add nim support for quicktype: for our custom usecase I decided it can be done quickly in a custom way, but it's very possible that this is my like-to-write-a-compiler-like thing bias

PR #342, opened 2025-08-19
Generate rust dap types from a single source of truth (an original json schema)

alehander92:
1) ~done: generate rust types from official dap JSON schema from https://github.com/microsoft/vscode-debugadapter-node

also base for TODO-s in future PR-s:

2) generate ct extension types JSON schema from our rust types; now working for some types, maybe in a future PR for all that we need; 

3) TODO: nim generator: now we have a base generator from the rust original one, but we need to add a nim generator with overriden nim-specific logic; and use it to generate dap and ct types

resolving initial part of #341


Ro6afF (approved):
LGTM

PR #346, opened 2025-08-20
Integrate updated ruby recorder

SersemPeca:


src/ct/db_backend_record.nim
```diff
@@ -71,7 +71,7 @@ proc recordDb(
 
   let startArgs = case lang:
     of LangRubyDb:
-      @[rubyTracerPath, program]
+      @["--out-dir", fmt"{traceFolder}", program]
```
alehander92: just `traceFolder` not `fmt"{traceFolder}"`

src/ct/db_backend_record.nim
```diff
@@ -134,6 +134,8 @@ proc recordDb(
         # for other languages, we must start in the real inherited
         # work dir
         getCurrentDir()
+  echo "ABOUT TO START: ", vmExe, " WITH ARGS: ", startArgs.concat(args)
```
alehander92: remove or comment out before merge

alehander92 (approved):
ok: introducing probably back a regression, as we haven't fixed/improved the prints behavior in those ruby recorder versions: i hope it's ok

PR #352, opened 2025-08-25
fix: fix a lang detection bug and improve lang detection logic

alehander92:
(alexander): the bug was that when we had a `.` in a parent-parent folder of `program` on Nikola's machine, we were detecting the next part of the path as an extension, and also returning a false LangC default
(`ct record` wasn't working for `ct record .` in such a noir folder)

overally the logic was messy, refactored a bit according to Nikola's suggestions


alehander92:
@NGPetrov tested this on his machine, IIRC it fixes the problem

PR #359, opened 2025-08-27
Refactor index.nim

Madman10K:
TODO:

- [x] Remove dead code from `index.nim`
- [x] Split index.nim into smaller files
- [x] Look into refactoring `index_config.nim` as well


src/frontend/index/tabs.nim
```diff
@@ -1,8 +1,11 @@
 import
```
alehander92: maybe this can be unified with files.nim: the function naming is just historical, but they usually work with the same things here: sources/files shown in tabs or other places

src/frontend/index/traces.nim
```diff
@@ -1,14 +1,115 @@
...
-  files,
-  ../../../common/[ ct_logging, paths ]
+  ipc_types/[ dap, socket ],
+  ../[ trace_metadata, config, types, lib ],
+  ../../common/[ ct_logging, paths ]
+
+
+var childProcessExec* {.importcpp: "helpers.childProcessExec(#, #)".}: proc(cmd: cstring, options: js = jsUndefined): Future[(cstring, cstring, js)]
+
+proc loadFilenames*(paths: seq[cstring], traceFolder: cstring, selfContained: bool): Future[seq[string]] {.async.} =
```
alehander92: TODO in this PR: i'd expect this to be in files.nim as well (and gradually moved out to the backend)

src/frontend/index/traces.nim
```diff
@@ -1,14 +1,115 @@
...
+        warnPrint "loadFilenames for self contained trace trying to read ", tracePathsPath, ":", err
+
+      for path, _ in pathSet:
+        res.add($path)
+    else:
+      # leave res empty
+      discard
+  return res
+
+proc loadSymbols(traceFolder: cstring): Future[seq[Symbol]] {.async.} =
```
alehander92, line 9: just a note in this case: those should be probably also moved out to backend soon (or in the future )

src/frontend/index/traces.nim
```diff
@@ -1,14 +1,115 @@
...
+    let (rawSymbols, err) = await fsReadFileWithErr(cstring(symbolsPath))
+    if err.isNil:
+      return ($rawSymbols).parseJson.to(seq[Symbol])
+    else:
+      # leave pathSet empty
+      errorPrint "loadSymbols for self contained trace trying to read from ", symbolsPath, ": ", err
+      return cast[seq[Symbol]](@[])
+
+
+proc loadFunctions(path: cstring): Future[seq[Function]] {.async.} =
```
alehander92, line 21: just a note: those as well (not in this PR)

alehander92 (changes requested):
good work: hard to say how intuitive the new file structure would be, but we can iterate on that: thank you!

PR #360, opened 2025-08-27
Fix clash between local and system ct

Madman10K:
On Linux distributions where both both the ct appimage/package is installed and one tries to run ct locally from a dev shell the system ct is used. This should not be the case.


alehander92 (approved):
good

SersemPeca:
lgtm

Issue #337, opened 2025-08-19
Add `--diff` or `--with-diff` parameter to`ct record`


alehander92:
It can support adding the diff / patch file for a certain situation to the trace. We can pass for now special keywords like `last-commit` or maybe git-compatible expressions like `HEAD~..HEAD` or a commit hash? or maybe directly a patch file. The very first version can start with passing something like `last-commit` or the path to a patch file, as most useful for experiments

We might directly parse the diff file here, as it would be useful for indexing/postprocessing commands and for the frontend in a structured form

Issue #338, opened 2025-08-19
Diff view


alehander92:
depends on #337  or #348 

EDIT: PR: #376

PR #366, opened 2025-09-01
feat: start adding code review commands/multitrace support

alehander92:
start working on record --with-diff, but this can be reused in a command like `ct make-multitrace`

resolves https://github.com/metacraft-labs/codetracer/issues/337

for now supported in `ct replay`, which for now sends the structured diff data to frontend; :
```
ct record --with-diff=last-commit <some-program> # with-diff can also be a git range syntax, we just pass it to `git diff` for now; 
# also it can be a diff file, but then we MUST call `ct record` from its repo root folder for now
ct replay -t=<multitrace-archive>`
```
maybe in separate work `ct host`, `ct upload` ?


alehander92:
i hope it's ok: we will have further PR-s; (Nedy has worked on top of this and Franz tested it for regressions before me rebasing on 1 more commit in #338 )

PR #368, opened 2025-09-03
feat(frontend): terminate backend-manager on all app exits

Ro6afF:
- Add stopBackendManager() to centralize and idempotently kill the backend-manager process.
  - Hook cleanup into Electron lifecycle: before-quit and window-all-closed.
  - Handle Node process signals (SIGINT, SIGTERM, SIGHUP) to stop backend-manager and quit.
  - Add process exit cleanup as a final safeguard.
  - Replace ad-hoc kill in onCloseApp with centralized cleanup.
  - Update codebase insights to document the lifecycle and cleanup strategy.

### DIFF
--- a/.agents/codebase-insights.txt
+++ b/.agents/codebase-insights.txt
@@ -10,3 +10,10 @@
 
 
 - The db-backend DAP server now responds to the `configurationDone` request.
+
+- Electron lifecycle: the backend-manager process is spawned in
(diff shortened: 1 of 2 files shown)


alehander92 (approved):
sounds ok: not sure if we should exit with 0 on SIGINT, but it probably works

PR #374, opened 2025-09-11
Fix fatal crashes caused by incorrect use of temporary directories

Madman10K:
We cannot use `/tmp` on macOS as listed in previous issues.


src/backend-manager/src/paths.rs
```diff
@@ -0,0 +1,28 @@
+use std::env;
```
alehander92, line 1: it would be good to be able to reuse that between db-backend and backend-manager

PR #375, opened 2025-09-11
Synchronise the version in version.nim with the version in Info.plist

Madman10K:
Currently we're not synchronising versions between `version.nim` and `Info.plist`. This patch replaces with the correct version on each build.


alehander92 (approved):
thank you

PR #382, opened 2025-09-22
Fix macOS bundles and some tmp-related fixes

Madman10K:
Currently, only the tup and nix builds run without any issues.

TODO:

- [x] Fix macOS bundle
- [ ] Fix AppImage build


alehander92 (approved):
thanks ! good fixes

PR #386, opened 2025-09-30
Feat: `ct index-diff` and initial db backend impl

### LINKED ISSUE #343: Add `--index-diff` to `ct record` or optionally as a separate `ct index-diff`(and maybe `backend index-diff` command: for beginning for db-backend)

alehander92:
needed for #344 resolves #343

after pairing with @pxor also adding flow mode to `ct/load-flow`: can load diff flow for multitraces in dev mode: instead of a new `ct/load-diff-flow` . Resolving #344

### DIFF
--- a/src/common/common_types/codetracer_features/frontend.nim
+++ b/src/common/common_types/codetracer_features/frontend.nim
@@ -206,6 +206,7 @@ type
     rawTestStrategy*: langstring
     diff*: Diff
     withDiff*: bool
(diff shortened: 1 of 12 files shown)


alehander92:
paired a bit on it with @pxor : those commands/cases are still experimental, so hopefully ok to merge

@Franz-Fischbach didn't see obvious regressions in the normal `Call` flow mode

PR #250, opened 2025-07-14
docs: add stylus demo guide

Franz-Fischbach:
## Summary
- document how to run the Stylus demo in `stylus_and_wasm.md`

## Testing
- `cargo build`
- `cargo test`
- `cargo clippy`

------
https://chatgpt.com/codex/tasks/task_b_6874f4d87b60832a9bcf3358d0c72651

Issue #246, opened 2025-07-10
Current Codetracer.app release crashes on MacOS


tzanko-matev:
After running `xattr -c ...` command I get the following crash report. Note that I downloaded the app image from Github and didn't build it on my own.

-------------------------------------
Translated Report (Full Report Below)
-------------------------------------

Process:               ct [71059]
Path:                  /Applications/CodeTracer.app/Contents/MacOS/bin/ct
Identifier:            ct
Version:               ???
Code Type:             ARM-64 (Native)
Parent Process:        launchd [1]
User ID:               501

Date/Time:             2025-07-10 23:14:43.2584 +0300
OS Version:            macOS 14.6.1 (23G93)
Report Version:        12
Anonymous UUID:        849FC086-23C8-B5C0-7EF9-B9E0C390678E

Sleep/Wake UUID:       52EFC84C-9DA5-46C6-92DF-CD7DBC559F59

Time Awake Since Boot: 800000 seconds
Time Since Wake:       486 seconds

System Integrity Protection: disabled

Crashed Thread:        0

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000

Termination Reason:    Namespace DYLD, Code 1 Library missing
Library not loaded: /opt/homebrew/*/libzip.5.dylib
Referenced from: <83EC8A39-93CB-3D31-B3D3-7CA340B9C787> /Applications/CodeTracer.app/Contents/MacOS/bin/ct
Reason: tried: '/opt/homebrew/*/libzip.5.dylib' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/opt/homebrew/*/libzip.5.dylib' (no such file), '/opt/homebrew/*/libzip.5.dylib' (no such file)
(terminated at launch; ignore backtrace)

Thread 0 Crashed:
0   dyld                          	       0x19aa4b4c8 __abort_with_payload + 8
1   dyld                          	       0x19aa560cc abort_with_payload_wrapper_internal + 104
2   dyld                          	       0x19aa56100 abort_with_payload + 16
3   dyld                          	       0x19a9e77f0 dyld4::halt(char const*, dyld4::StructuredError const*) + 304
4   dyld                          	       0x19a9e42ec dyld4::prepare(dyld4::APIs&, dyld3::MachOAnalyzer const*) + 3888
5   dyld                          	       0x19a9e2ef4 start + 1868

[...]

PR #391, opened 2025-10-06
Resolve security vulnerabilities in some rust dependencies

Madman10K:
Should fix the following dependabot alerts:

#250 
#246
#245
#239 
#236 
#221

### DIFF
--- a/src/ct/trace/multitrace.nim
+++ b/src/ct/trace/multitrace.nim
@@ -16,7 +16,7 @@ proc findDiff(diffSpecification: string): string =
       except OsError:
         discard
 
-      if existsFile(path):
+      if fileExists(path):
         # assume it contains a diff
         # AND that we're in the actual repo as well
         readFile(path)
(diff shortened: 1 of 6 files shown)


src/ct/trace/multitrace.nim
```diff
@@ -16,7 +16,7 @@ proc findDiff(diffSpecification: string): string =
       except OsError:
         discard
 
-      if existsFile(path):
+      if fileExists(path):
```
alehander92, line 19: strange, i didn't realize both of those methods exist

alehander92 (approved):
good work

src/ct/trace/multitrace.nim
Madman10K: `existsFile` is deprecated in favour of `fileExists`. We get warnings about this from the nim compiler, so it's good to resolve it.

alehander92: ok, makes sense

PR #393, opened 2025-10-06
Cleanup packages/dependabot 2: try to remove some unused node packages to prevent dependabot alerts for them

alehander92:
tried to improve things with @Madman10K (his initiative)
TODO: problem with building (direnv reload leads to an error)


alehander92:
now it seems to build for me and @SersemPeca : if it has a problem again, we'll debug it

PR #406, opened 2025-10-13
Python integration

tzanko-matev:


src/ct/db_backend_record.nim
```diff
@@ -59,7 +59,7 @@ proc recordDb(
     lang: Lang, vmExe: string,
     program: string, args: seq[string],
     backend: string, traceFolder: string, stylusTrace: string,
-    traceId: int): Trace =
+    traceId: int, pythonActivationPath: string = "", pythonWithDiff: bool = false): Trace =
```
alehander92: `withDiff` should be typically lang-agnostic: currently its logic is implemented independently of the recorders, not sure if we need to pass it to them: even if we do, then it should be `withDiff`, not specifically `pythonWithDiff`

src/ct/db_backend_record.nim
```diff
@@ -233,14 +244,46 @@ proc record(
...
+        LangPythonDb,
+        interpreterPath,
+        executable,
+        args,
+        backend,
+        outputFolder,
+        stylusTrace,
+        traceId,
+        pythonActivationPath = activationPathResolved,
+        pythonWithDiff = pythonWithDiff)
```
alehander92: ditto: -> `withDiff` (?)

src/ct/db_backend_record.nim
tzanko-matev: Fixed

src/ct/db_backend_record.nim
tzanko-matev: Fixed

ci/test/python-recorder-smoke.sh
```diff
@@ -0,0 +1,149 @@
...
+set -euo pipefail
+
+ROOT_DIR=$(git rev-parse --show-toplevel)
+cd "${ROOT_DIR}"
+
+echo '###############################################################################'
+echo "Building Codetracer CLI for smoke test"
+echo '###############################################################################'
+
+./ci/build/dev.sh
```
SersemPeca: This script does not build the ct binary, hence this test will fail. We should run `just build-once` or something like that here so as to setup the `ct` binary.

PR #409, opened 2025-10-14
feat: ct launcher

Ro6afF:
Start splitting codetracer into multiple components and implement launcher to handle this splits.


src/ct/src/subcommands/record.rs
```diff
@@ -1,10 +1,21 @@
 use std::{path::PathBuf, str::FromStr};
 
-use crate::{Lang, RecordOptions};
+use crate::{db, subcommands::run_external, Lang, RecordOptions};
 
 pub fn run_record(options: RecordOptions) {
+    println!("{:#?}", options);
```
alehander92, line 6: we should return `Result` for most of the internal functions

src/ct/src/subcommands/record.rs
```diff
@@ -1,10 +1,21 @@
...
+
     let lang = options.lang.or_else(|| detect_language(&options.program));
-    println!("{:#?}", lang);
+    if lang.is_none() {
+        panic!("Can't determine language"); // TODO: better error and what to do?
```
alehander92, line 10: return `Err("..".into())` (or a more specific kind of error)

alehander92 (changes requested):

PR #418, opened 2025-10-21
Appimage refactor

tzanko-matev:
We've moved a lot of the logic of the old build_appimage.sh script to nix derivations. We've retained only the impure parts of fetching node modules and electron. 

One can review this PR by going over the commits one-by-one. Each commit produces a valid AppImage by running `./appimage-scripts/build_appimage2.sh`


Madman10K:
@tzanko-matev the best way to do it would be to test locally on these systems. You just need to download the AppImage on these platforms, make it runnable with `chmod +x`, run it and test it.

Link to the latest AppImage built by this PR: <https://downloads.codetracer.com/CodeTracer-418/merge-amd64.AppImage>

I'll probably do the testing myself later today though

Madman10K (approved):
Confirmed to work correctly on all target platforms.

SersemPeca (approved):

node-packages/yarn-project.nix
```diff
@@ -74,105 +52,96 @@ let
       rm $out/.gitignore
     '';
     outputHashMode = "recursive";
-    outputHash = (
-      if stdenv.isAarch64 then
-        "sha512-+f3z4zwF2YlzskL7uDwVvQ3Fg0EYJ1HGPEqIGNb2VsiQpXoY2tU+EufqX79YRCOW7lXkuDfMQHYI3XcPioAEvg=="
-      else
-        "sha512-8spVLTYEciHn5q0N1R4fRB9EBM4gzZ1Zd4EeZdLwQM6ob+wvgP3oMH6JHQXUxKraP8TlpuVg7l2gNsTaAH5NoQ=="
-    );
+    outputHash = "sha512-8spVLTYEciHn5q0N1R4fRB9EBM4gzZ1Zd4EeZdLwQM6ob+wvgP3oMH6JHQXUxKraP8TlpuVg7l2gNsTaAH5NoQ==";
```
alehander92, line 55: we need to do a dry build of the yarn cache for arm (it can be done on our architecture as well) and keep the `if` here with the new hash

Madman10K: Good luck with that 😂

I suggest we drop support for arm(this hash is for arm for Linux), since we don't test for it, nor do we have record of it building successfully there or of us having any users.

Computing the hash is a nightmare and we never found any instructions on how to actually get it. For the record, running a dry run does not compute the hash in any way. 

I'm in support of merging as is due to us losing up to 12 days on this previously(2 times in february, 4 times during april, 3 times during may, 2 times in june)

alehander92: ask @SersemPeca if he is ok with that; my impression is @zah wants the arm build; but i also don't know if we test the nix-based builds on arm or if anyone has tried them on arm(osx build doesn't use nix for now); i also agree that we can re-add the hash when we actually start to build/test there, but not sure

PR #419, opened 2025-10-21
Pip package for ct utility

Ro6afF:


build-python/setup.cfg
```diff
@@ -0,0 +1,45 @@
+[metadata]
+name = codetracer
+version = 25.9.2
```
alehander92, line 3: do we want this version schema(calendar) for pip? probably yes, but to make sure?

tzanko-matev: it's fine

PR #423, opened 2025-10-23
Try to update yarn dependencies & fix other security warnings

Madman10K:


src/db-backend/src/c_compat.rs
```diff
@@ -5,8 +5,6 @@ use core::ffi::{c_char, c_int, c_void};
 use core::mem::{align_of, size_of};
 use core::ptr::null_mut;
 
-#[global_allocator]
-static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
```
SersemPeca: Why do we remove this ? We need it for malloc in wasm. We only load it when the wasm feature is set.

PR #431, opened 2025-10-29
feat: ct remote command

Ro6afF:


nix/packages/default.nix
```diff
@@ -244,6 +244,40 @@
           };
         };
 
+        ctRemote = stdenv.mkDerivation rec {
+          pname = "ct-remote";
+          version = "03c01ce";
```
KvorHPamuL: this is not the final version! (need to set the domain + confirm the cli arguments)

PR #451, opened 2025-11-07
Fix macOS builds requiring Rosetta by default and create new release

Madman10K:


release_checklist.md
```diff
@@ -5,7 +5,7 @@
 * Update version.nim with the new version, following [calendar versioning](https://calver.org/): `YY.OM.<build>`:
   e.g. the first build for a month can be `25.03.1`, the next one `25.03.2` etc
 * Update Info.plist with the new version following the same format.
-* Update `build-python/setup.cfg` with the new version following the same format.
+* Update `build-python/setup.cfg`(if there are python changes) with the new version following the same format.
```
alehander92: we need to always update it so we include the newer `ct` in a package with the same version

PR #452, opened 2025-11-07
feat: Implement switching to editor mode for monaco and lsp support

SersemPeca:


src/frontend/lsp_controller.js
alehander92, line 1: probably by mistake: remove from commit/add gitignore rule (?)

alehander92 (approved):
thank you!!

PR #463, opened 2025-11-19
`ct host` timeouts and frontend notifications

tzanko-matev:
We've implemented a timeout for `ct host` - the process will exit with code 0 after some time without receiving any messages from the frontend. The default timeout is 10 minutes. It can be configured by `ct host --idle-timeout=xxx`. This makes it possible for the CI product to automatically stop old unused replay sessions.

This also makes the replay sessions list in https://web.codetracer.com unnecessary

Also we've added a notification in the frontend when the backend is no longer connected to it. This can happen if the backend exits or if we open another browser tab and connect to the backend - the old frontend will no longer work.

[Screencast From 2025-11-19 13-26-16.webm](https://github.com/user-attachments/assets/93b6ef66-d3b1-4097-9666-73e637db5fce)


src/ct/trace/host_idle_timeout_test
alehander92, line 1: please remove(added by mistake)

src/frontend/tests/idle_timeout_test
alehander92, line 1: please remove(added by mistake)

PR #468, opened 2025-11-25
Fix re-record functionality for noir & context menu + flow after monaco upgrade

SersemPeca:


src/frontend/renderer.nim
```diff
@@ -1286,7 +1286,22 @@ proc reRecordCurrentTrace*(data: Data) =
     data.viewsApi.errorMessage(cstring"Current trace does not define a program to run.")
     return
 
-  var args: seq[cstring] = @[data.trace.program]
+  var programArg = data.trace.program
+  if data.trace.lang == LangNoir and data.trace.workdir.len > 0:
+    # Noir metadata stores the project name; re-record requires the project root.
+    programArg = data.trace.workdir
+
+  # Some recorders (e.g. Noir) persist only the project name in metadata.
```
alehander92, line 1293: we maybe need to add this to metadata or to have somewhere the mapping but for now this is ok!

PR #480, opened 2025-12-01
feat: Begin implementing acp wiring and UI elements

SersemPeca:


src/ct/acp/opencode_demo.js
alehander92, line 1: remove this if artifact

src/frontend/index/ipc_subsystems/acp_ipc.nim
```diff
@@ -0,0 +1,120 @@
...
+  let stream = ndJsonStream(
+    toWebWritable(stdinOf(procHandle)),
+    toWebReadable(stdoutOf(procHandle)))
+
+  echo "[acp_ipc] set up the pipes"
+
+  var aggregatedContent = cstring""
+  var collectedUpdates: seq[JsObject] = @[]
+
+  let handleSessionUpdate = functionAsJS(proc(params: JsObject) {.async.} =
```
alehander92: maybe this can be typed? but if its' very dynamic ok

PR #489, opened 2025-12-05
feat(tracepoint): implement field access

Ro6afF:


alehander92 (approved):
thank you!

in the future we might have a special `loadField/loadExpression` for more optimal field evaluation for rr; for now great!

PR #492, opened 2025-12-08
feat: Initial multi agent support

pxor:


src/frontend/ui/agent_activity.nim
```diff
@@ -126,7 +126,7 @@ proc sendAcpPrompt(prompt: cstring) =
 
 proc submitPrompt(self: AgentActivityComponent) =
   # Use the latest input value to avoid stale data.
-  let inputEl = document.getElementById(INPUT_ID)
+  let inputEl = document.getElementById(INPUT_ID & fmt"-{self.id}")
```
SersemPeca: I don't see the AgentActivityComponent type having an `id` field. Also, I think I'd be a good idea to have the AgentComponent's id be the id of the session that "belongs" to it. So maybe we ought to define a ``sessionId`` variable in the AgentActivityComponent and modify the code to call ``self.sessionId`` instead of ``self.id``

SersemPeca: Resolved: This id is automatically generated, so we'll fall back to it and keep the sessionId as a separate field

PR #499, opened 2025-12-16
by Dimo: feat(flow/rust): ignore some identifiers that cannot be variables:don't show them in the flow

alehander92:
by @Ro6afF


alehander92 (commented):
good! (original commit is by Dimo: approving , github doesn't let me do it because I have opened the PR)

PR #502, opened 2025-12-18
Install and use the CodeTracer RR back-end

zah:


quicksort
alehander92, line 1: sorry: please remove binary file

src/common/config.nim
```diff
@@ -153,6 +153,14 @@ proc loadConfig*(folder: string, inTest: bool): Config =
     var c = Config()
     c[] = config
     c.shortcutMap = initShortcutMap(config.bindings)
+
+    # Auto-discover ct-rr-support from PATH if not configured
+    if c.rrBackend.path == "":
```
alehander92, line 158: shouldn't PATH have priority?

alehander92: no, that's correct, nvm

PR #516, opened 2025-12-23
Nim tree-sitter integration and omniscience support

zah:
Major changes:

* Alias the nim compiler used for building to 'nim1' (this frees up the nim compiler used for recording to be a diffrent version)
* Build several versions of Nim for testing
* Create an integration test that test each of them
* Fix compatibility issues with the varied Nim versions, discovered during testing
* Introduce a similar setup for Rust


src/db-backend/tests/test_harness/mod.rs
```diff
@@ -0,0 +1,664 @@
...
+use std::env;
+use std::fs;
+use std::io::{self, BufRead, BufReader, ErrorKind, Write};
+use std::os::unix::net::{UnixListener, UnixStream};
+use std::path::{Path, PathBuf};
+use std::process::{Child, Command, Stdio};
+use std::time::{Duration, Instant};
+use std::thread;
+
+/// Supported languages for test programs
```
alehander92, line 29: we do have an existing `Lang` enum, and extension helpers for it somewhere, really better to reuse it in the future: probably not a problem for the test, just a general note

alehander92: `Lang` is in lang.rs, we have also `lang_from_context` (path + trace kind) , but the extension thing is hardcoded for now in flow_preloader(and we have helpers in nim, not rust for it, sorry)

PR #622, opened 2026-08-13
fix(ct-test): recover the Nim tests lost after a numeric type suffix

zah:
## What

`ct test discover` was losing Nim test cases — and, less visibly, inventing a few.

The `nim-unittest` provider decided what was code and what was a literal with two
hand-rolled character loops, and both treated every apostrophe as the opening quote
of a character literal. Nim gives `'` three other jobs: the type suffix on a numeric
literal (`0'u8`, `1.0'f32`, `0x1F'i64`), the custom numeric literal (`12'MyType`),
and accent-quoted operator names (`` `'u8` ``). One `sizeBytes = 10485760'i64`
desynchronised the scan, and everything between there and the next apostrophe —
usually the rest of the file — was skipped as string content. Those files reported

```
std/unittest import detected but no literal suite/test declarations were found
```

while containing perfectly ordinary `suite "…":` / `test "…":` at column 0.

## How

Both scans now share one lexer, `src/ct_test/nim_lexer.nim`.

It is a new module rather than a reused one, and the header says why: the Nim
distribution's `compiler/lexer.nim` is the ground truth every rule here is written
against, but it is not a package dependency, it lives only inside the Nim
installation at a store path that moves with the toolchain, it drags in the
compiler's global configuration, and `lexMessage` terminates the process on a
malformed literal — which is disqualifying for a scanner that reads untrusted
third-party sources.

Covered: numeric type suffixes (with and without the apostrophe), character
literals including `'\''` and `'\x41'`, raw strings (`r"C:\dir\"` — no backslash
escapes), generalized raw string literals (`re"a""b"`), triple-quoted strings with
their `"""` -not-followed-by-`"` terminator rule, and nested block comments in both
the `#[ ]#` and `##[ ]##` forms.

Fail-safe by construction: anything Nim requires to close on its own line ends at
the newline, and an apostrophe with no plausible closer degrades to one ordinary
character. A malformed file costs one mis-read token, never the remainder of the
file.

## Effect

Measured across sixteen Nim repositories with the pre- and post-fix binaries:

| | before | after |
| --- | ---: | ---: |
| cases | 10,305 | **10,901** |
| suites | 2,468 | **2,602** |
| files whose count changed | — | 115 |
| files warned "no literal suite/test declarations" | 24 | **1** |

The surviving warning is a helper module that imports `std/unittest` and declares
nothing — a true positive.

Sample, per file:

| file | cases before → after |
| --- | --- |
| `codetracer/src/tests/gui/tests/views/isonim_views_test.nim` | 398 → 479 |
| `codetracer/src/frontend/viewmodel/tests/unit/test_collab_reducer.nim` | 3 → 14 |
| `isonim/tests/test_hydration.nim` | 1 → 20 |
| `reprobuild/tests/e2e/local-build-engine/t_e2e_local_reprobuild_project_build.nim` | 0 → 18 |
| `reprobuild/libs/repro_dsl_stdlib/tests/t_c1_catalog_round_trip.nim` | 0 → 6 |
| `io-mon/tests/portable/test_io_mon_t0_completeness.nim` | 40 → 55 |

Every discovered item was cross-checked against the source: all ~10,900 sit at the
start of their line with the expected keyword, and an independently written scanner
finds no declaration that discovery misses.

Discovery wall time is unchanged: 5.8s before, 5.6s after, on the largest workspace
measured.

## The mirror defect

The same audit found discovery **over**-counting: `suite`/`test` inside a
`when false:` block were reported although the compiler never instantiates them, so
no runner can produce those cases. They are now skipped — with an `info` diagnostic
naming the count, not silently. (No over-counting was found from comments, strings
or here-docs; those were already handled and are now pinned.)

## Also in this change

- `CapturedRun` gains `outputBytes` and `truncated`. `execCaptured` bounds captured
  output at 16 MiB and used to drop the fact that it had cut anything, leaving every
  consumer to parse a prefix as if it were the whole answer. For a JSON or XML
  payload that is a loud parse error; for the list-shaped outputs ct_test consumes it
  is invisible. Workspace scoping now refuses a truncated `git ls-files -z` inventory
  outright — git emits paths sorted, so truncation removes a contiguous tail of the
  alphabet — with `--scope auto` falling back to the pruning filesystem walk (and
  saying why) and `--scope vcs` reporting the scope unavailable.
- `--scope` and `--unscoped` now appear in the usage string, and the diagnostic for
  an unusable inventory names the surface the mode actually came from instead of
  always blaming `CT_TEST_SCOPE`.
- The note claiming nested-repository boundaries are what exclude vendored
  `references/` trees is corrected: in the measured workspace those paths are
  gitignored and `git ls-files` reported no directory entries, so the ignore chain
  did all of the work. The nested-checkout rule is the backstop for a workspace that
  vendors a tree *without* ignoring it.

## Tests

- `src/ct_test/nim_lexer_test.nim` — 21 cases, one per construct.
- `ct-test Nim unittest lexical regressions` in `nim_unittest_provider_test.nim` —
  file-in/catalog-out, no mocks.
- `ct-test workspace scoping guards` in `discovery_test.nim` — the truncation guard
  is exercised by lowering the capture bound, not by materialising 280,000 files.
- `fixtures/nim_unittest_project/tests/test_lexical_edge_cases.nim` — a **compilable**
  unittest file, so `nim c -r` on it is an independent ground truth for what
  discovery must report (1 suite + 5 cases; discovery agrees exactly).

Mutation-verified: 13 mutations applied one at a time, each confirmed to make the
suite that claims to pin it fail, then reverted. Three of the first twelve initially
passed and exposed real coverage gaps (the `"""`-terminator rule, the character
lookahead bound, and unterminated name strings); tests were added and all three now
fail under mutation. A thirteenth restores the original scanner's algorithm wholesale
and fails both the lexer and the provider suite.

`ci/test/m16-release-gate.sh` passes end to end locally (133 assertions, exit 0).

---

## Merge evidence (2026-08-14)

Merged on `ct-test-release-gate` alone, per owner direction. Stating plainly what was and was not observed:

**Observed (CI):** `ct-test release gate (M16 provider matrix)` — **pass**, 16m49s, [run 31706159237](https://github.com/metacraft-labs/codetracer/actions/runs/31706159237/job/94467134706).

**Not observed:** `ct-test-providers` (the "ct-test cross-language provider gate") — failed in 2m48s in the same run, in `Setup dev env`, as part of the workspace-lock outage. It never reached this provider and says nothing about this change either way. Most of the other red checks on this PR share that cause; PR #623 fixes it and this branch has been rebased onto the repaired base.

**Observed (local, not a CI result):** the three suites covering this change, run from the repo root in the dev shell —

| suite | result |
| --- | --- |
| `src/ct_test/nim_lexer_test.nim` | pass, 21 assertions |
| `src/ct_test/nim_unittest_provider_test.nim` | pass, 19 assertions |
| `src/ct_test/discovery_test.nim` | pass, 13 assertions |

These are local runs and are **not** presented as CI results.

### Outstanding obligation

`ct-test-providers` **must be observed green before the next change that touches `src/ct_test/frameworks/nim_unittest.nim` or `src/ct_test/nim_lexer.nim`.** A passing `ct-test-release-gate` does not cover it — the two gates have deliberately different prerequisites, and only the providers gate drives the real recorder siblings.

This obligation is recorded in the provider source itself, not only here, so the next person editing the file sees it.


zah:
### CI triage — every failure so far is pre-existing on `dev`

Recording this in-band so nobody has to re-derive it. Each red check on this PR has been
matched to the same failure on `dev` *before this branch existed*:

| check | failing step | pre-existing evidence |
| --- | --- | --- |
| `visual-replay-regression-gate` | `Setup dev env` | same step on `dev` @ `253a4f67` (run 31551743085) |
| `test-ui-tests (macos, macos-latest)` | `Setup dev env` | same step, same run |
| `test-non-gui (macos, macos-latest)` | `Build (non-nix)` | byte-identical error on `dev` runs 31547476316 / 31540196342 / 31537755930 |

The third one deserves a note, because the baseline run I first compared against had
failed *earlier* (at `Checkout`) and therefore proved nothing about `Build (non-nix)`.
Checking further back, the error is verbatim:

```
ERROR: codetracer-trace-format checkout not found at .../non-nix-build/../../codetracer-trace-format.
  Clone it as a workspace sibling, or set TRACE_FORMAT_DIR to point at an existing checkout.
```

on `dev` at 2026-08-11, two days before this branch. It is a missing workspace sibling in
the macOS non-nix lane. This PR touches `src/ct_test/*`, one line of
`ci/test/m16-release-gate.sh`, and a notes file — no build wiring, no sibling resolution,
no Rust/FFI.

**The two jobs that actually cover this change — `ct-test-release-gate` and
`ct-test-providers` — have not started.** The `garm-*` self-hosted pool is saturated and
the whole run (including `dev`'s own) has been `queued` for over an hour. Both are listed
in `ci/verdict/required-jobs.txt`, so their result is the thing worth waiting for here.

Locally, `ci/test/m16-release-gate.sh` passes end to end: 133 assertions, exit 0.

Not merging until those two land. `ci-verdict` currently fails on `dev` itself, so "all
green" is not presently attainable in this repo; that is a reason to wait for the relevant
gates and say so, not a licence to force the merge past them.

PR #625, opened 2026-08-14
fix(ci): check the private-Cargo credential boundary at the host, not the repo

zah:
## What

The visual-replay gate reaches a private Cargo git dependency using the organisation's CI App installation token — the same token that authenticates the job's own checkout and every private sibling the gate clones. The preflight nevertheless asserted, as `auth-header-credential-leak`, that this credential must not appear in any header applying to another `github.com` URL.

There is no narrower credential for a wider header to leak, so that assertion failed exactly when the job was provisioned correctly, and would have passed only if a second, single-purpose token had been introduced.

## Why the Actions-provided token is not an option

- *"The token's permissions are limited to the repository that contains your workflow"* — [GITHUB_TOKEN](https://docs.github.com/en/actions/concepts/security/github_token)
- *"`${{ github.token }}` is scoped to the current repository, so if you want to checkout a different repository that is private you will need to provide your own PAT"* — [actions/checkout](https://github.com/actions/checkout#checkout-multiple-repos-private)

One credential covering this repository and the private dependency is therefore the mechanism working, not a leak.

## What replaces it

The header stays narrowly keyed even though the credential is org-wide, and an org-scoped token handed to a host that is not `github.com` is a leak under any arrangement.

- The URL-scope reading now also probes off-host URLs — an unrelated host, a suffix lookalike of `github.com`, and the same path on another forge — so a host-less `http.extraHeader`, which Git applies to every URL it fetches, is caught rather than only a widening within `github.com`.
- The environment reading becomes `auth-header-offsite-leak`, checking those off-host URLs only, and keeps the credential-containment test that catches all four wire-valid spellings of one leaked header.

Everything else the preflight guards is untouched: no credential in `git-remote-https` argv, none persisted to disk, none printed on failure, no URL rewrite, no redirects, TLS verified, one pinned git source in the lockfile.

## Tests

`ci/test/visual-replay-private-cargo-preflight-test.sh` gains the case that would have caught this — the gate's own token applied `github.com`-wide by ambient config must **pass** — plus an off-host leak case and a host-less-key case. Each new assertion was mutation-checked: reverting the corresponding production change reddens exactly the intended case.

The suite runs on a stock runner in the verdict job as well as inside the GPU gate.

## Policy

Written up in `metacraft-dev-guidelines/policies/ci-workflow-standards.md` § *Which credential for which repo*, and the source carries a note saying why the retired assertion must not be reinstated.

### DIFF
--- a/.github/workflows/codetracer.yml
+++ b/.github/workflows/codetracer.yml
@@ -3033,11 +3033,17 @@ jobs:
           submodules: recursive
           token: ${{ steps.ci_token.outputs.token }}
           # Do not leave `http.https://github.com/.extraHeader` behind in
-          # `.git/config`. That persisted header prefix-matches every URL, which
-          # is what forced `visual-replay-private-cargo-preflight.sh` to weaken
-          # its `auth-header-url-boundary` check from "no header matches at all"
-          # to "the lldb-sys credential does not widen"; see the long comment
-          # there. Nothing downstream needs it:
+          # `.git/config` for the rest of a three-hour job on a persistent GPU
+          # runner. This is ordinary least-persistence hygiene — the same reason
+          # actions/checkout v6 moved the credential out of `.git/config`
+          # entirely — and NOT a boundary the private-Cargo preflight depends
+          # on: that header carries the org-wide CI App token, which by policy
+          # also authenticates the private lldb-sys dependency, so its presence
+          # is never a leak and the preflight no longer asserts anything about
+          # it. See the long note in
+          # `ci/test/visual-replay-private-cargo-preflight.sh`.
+          #
+          # Nothing downstream needs the persisted header:
           #   * `Setup dev env` gets the same token explicitly via `gh-token`.
           #   * `submodules: recursive` is unaffected — actions/checkout fetches
           #     submodules under a temporary *global* auth config it installs
--- a/ci/test/visual-replay-private-cargo-preflight-test.sh
+++ b/ci/test/visual-replay-private-cargo-preflight-test.sh
@@ -163,6 +163,19 @@ run_negative_case() {
 		# and it is the case that reddens if that reading is deleted.
 		GIT_CONFIG_KEY_0="http.https://github.com/metacraft-labs/.extraHeader"
 		;;
+	auth-key-hostless)
+		# The widening that matters most now the credential is known to be
+		# org-wide: a key with no URL section at all. Git applies
+		# `http.extraHeader` to *every* URL it fetches, github.com or not,
+		# so this would hand an organisation-scoped token to whatever host a
+		# dependency happens to name. Reading (1) catches it — via the
+		# same-host probes as well as the off-host ones, since a host-less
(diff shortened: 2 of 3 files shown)


zah:
Rebased onto current `dev` (`663b6ce09`) and added a second commit fixing a regression this PR introduced.

### The regression

CI reported one PR-introduced failure: `ci-verdict` → *Self-test the private-registry credential preflight*, on the new `auth-key-hostless` case. (`effective-auth-header` in that output is the invariant that *fired*, not a second failing case.)

It reproduced byte-for-byte locally once the runner's distinguishing feature was reproduced — a checkout carrying an ambient `http.https://github.com/.extraheader`:

```
Private Cargo preflight negative case failed: auth-key-hostless
Visual replay CI private Cargo authentication invariant failed: effective-auth-header.
```

**Cause: check ordering, surfaced only where an ambient header exists.** A host-less `http.extraHeader` is the least specific key there is, so against the github.com-wide header `actions/checkout` persists it loses Git's specificity contest — the lldb-sys URL resolves to the *ambient* header, `effective-auth-header` fires, and the isolated boundary probe is never reached. On a laptop, with no ambient header, the same key reaches the probe as intended.

The probe's host-less coverage was therefore live in development and dead in the one environment the gate runs in — the exact hazard the source already warns about for the literal-key check ("ordered the other way the literal check subsumed this one entirely and nothing could ever reach here"). `effective-auth-header` reintroduced it.

CI corroborates this independently: the very same self-test **passed** inside `visual-replay-regression-gate` on the broken commit, because that job is the one checkout in the tree with `persist-credentials: false` — no ambient header, no bug.

### The fix

Run the isolated probe first. Both orderings fail the gate, so nothing is relaxed; what changes is that the environment-independent reading answers first and the probe stays reachable. Plus two test changes:

* `auth-key-hostless` now runs from a checkout carrying the ambient header, so it reproduces the runner rather than the developer machine;
* a new `auth-key-narrow` case pins `effective-auth-header` itself — reordering had left it with no coverage at all.

### Mutation verification

Every check in the preflight was deleted or gutted in turn, in **both** environments, with a diff-proof that each mutant actually applied (two no-op mutants were caught and redone):

| mutation | caught by |
|---|---|
| revert the reorder | `auth-key-hostless` — now red in *both* environments, previously red in neither locally |
| gut isolated probe loop (reading 1) | `auth-key-wide` |
| delete `effective-auth-header` | `auth-key-narrow` (new) |
| gut off-host loop (reading 2) | offsite-leak cases |
| delete literal-key check | `auth-key` |
| delete probe-isolation guard | `probe-isolation` |

`shellcheck` clean; `shfmt -i 0` clean.

> **Note on the CI signal:** `ci-verdict` waits on jobs whose `runs-on` label has no registered runner, so it publishes ~24h after a push (verified: the previous run's macOS jobs were cancelled at exactly +24h and `ci-verdict` started 2s later). This PR should not be merged until that verdict lands.

PR #627, opened 2026-08-16
fix(ci): read a routed lock record as an unlocked commit, not a broken one

zah:
A routed per-repo lock record made the sibling-revision resolver abort the job
instead of degrading like a missing lock.

## What is wrong

`repro locking adopt-manifest` puts a workspace into **routed** locking mode.
In that mode the workspace tooling stops writing the monolithic workspace-lock
document and instead writes one minimal per-repo record — into the **same**
`locks/<project>/<repo>/<sha>.toml` namespace the lock documents use:

```toml
[[repo]]
name = "codetracer"
path = "codetracer"
revision = "<sha>"
```

That is not a workspace lock. It declares no schema and pins only the repo
whose directory it sits in, so it can never say what revision a *sibling* is
at. Read as a lock it fails with `no top-level 'schema' key` and **exit 5** —
the code meaning "a lock exists but cannot be trusted", which callers escalate
to a job abort. The honest answer is the far milder "this commit has no
workspace lock" (**exit 3**), which every caller already degrades from
gracefully.

So a commit that merely *lacked* a lock was killing the job. Reproduce on the
published data:

```
scripts/resolve-sibling-rev.sh --repo codetracer --sibling codetracer-nim \
  --sha 81f8e91f0aa882bff85c99bf8a4edb2ac0caff40 --no-walk   # was exit 5, now exit 3
```

## Scale, measured

Against a fresh clone of `metacraft-labs/metacraft-manifests@latest`:

| | count |
|---|---|
| `.toml` files under `locks/` | 127 |
| genuine `reprobuild.workspace.lock.v1` documents | 29 |
| **routed per-repo records (no `schema` key)** | **98**, covering **96 repos** |
| of those, reproducing **exit 5** today | **71** |

The other 27 already resolved because a valid `.xml` lock for the same commit
won the project-preference tie-break. After this change those 71 become 68
graceful exit 3, 2 **successful resolves** (the record had been poisoning a
real lock written for the same commit), and 1 honest exit 4.

These records are already published, so this cannot be fixed by changing what
the tooling writes next.

## The fix

Recognise the records at **discovery** and exclude them from the lock set.
That beats parsing them into an exit 4: a commit whose only record is routed
now falls through to the **next candidate SHA**, and the ancestry walk still
reaches the locked parent — exactly as an unlocked commit does.

Recognition is **semantic**, not a syntax fingerprint, so adding a field to the
record cannot silently restore the abort:

* the document declares no top-level `schema`, **and**
* every table in it is `[[repo]]`, **and**
* the only repo name it pins is the one being resolved.

A schema-less document naming some *other* repo is still refused loudly — it
claims to know a sibling's revision without declaring what it is — and so is
one carrying a non-`[[repo]]` table. The exit-3 diagnostic names every record
it ignored and why, so the degradation is never silent.

## Tests

Ten contracts added to the shared suite (76 → 86 assertions). Each was
mutation-verified:

* all 7 behavioural contracts fail against the unfixed resolver;
* dropping the "names nobody but itself" condition kills the over-breadth guard
  **and** two pre-existing contracts;
* dropping the "only `[[repo]]` tables" condition, the top-level-schema escape,
  and the diagnostic each kill exactly their own guard.

No pre-existing contract changed.

## Companion

`metacraft-labs/metacraft-github-actions#<pr>` carries the identical change to
`clone-siblings/resolve-sibling-rev.sh` — the copy CI actually executes. This
file is the vendored copy and the two are kept byte-identical below the header.

### DIFF
--- a/scripts/resolve-sibling-rev-test.sh
+++ b/scripts/resolve-sibling-rev-test.sh
@@ -32,7 +32,7 @@ if [[ ! -x $RESOLVER && ! -f $RESOLVER ]]; then
 	exit 3
 fi
 
-EXPECTED_ASSERTIONS=76
+EXPECTED_ASSERTIONS=86
 
 PASS=0
 FAIL=0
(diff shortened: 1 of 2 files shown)


zah:
Companion PR (the copy CI actually executes): metacraft-labs/metacraft-github-actions#6

PR #632, opened 2026-08-16
fix(ci): resolve the MCR DAP flow's siblings the way every other job does

zah:
`mcr-dap-flow` resolved its four sibling revisions by grepping
`.github/sibling-pins`. That file was deleted eight weeks ago, when every
cross-repo CI site in this repo was migrated to resolve sibling revisions
from the workspace lock instead of from repo-local pin files. This
workflow was the single site that migration missed.

Since then the job could not have succeeded. `grep` on a file that does
not exist exits 2, and the reads sit in command substitutions under
`set -euo pipefail`, so the job died in "Resolve sibling refs" before
cloning a single sibling. It went unnoticed because the workflow is
path-filtered onto `src/db-backend/**` and `libs/ct-dap-client/**` and so
fires rarely -- the same reason the checkout-ordering defect in
`beam-flow.yml` survived as long as it did.

Re-creating the pin file would reintroduce exactly what that migration
removed, so the resolution now goes through `setup-dev-env`'s `siblings:`
list. That is the shape this workflow's own header already says it
mirrors from `beam-flow.yml`, and it collapses "Install Nix", "Resolve
sibling refs" and four `clone-repo` steps into one step that installs Nix
and clones all four siblings adjacent to the checkout.

The refs are spelled `=dev` rather than left bare, matching every other
job here: a bare entry resolves through the per-commit workspace lock in
the shared manifests repo, which has published nothing for this repo
since 2026-08-02, and every job that left it bare died in that step. The
`workflow_dispatch` overrides for the two recorder/backend refs are
preserved; their descriptions no longer point at the deleted file.

The two `paths:` filters also listed the deleted file, so a workflow that
could only fail was additionally being triggered by a path that can never
change again.

### DIFF
--- a/.github/workflows/mcr-dap-flow.yml
+++ b/.github/workflows/mcr-dap-flow.yml
@@ -22,7 +22,6 @@ on:
       - 'libs/ct-dap-client/**'
       - 'scripts/detect-siblings.sh'
       - 'justfile'
-      - '.github/sibling-pins'
       - '.github/workflows/mcr-dap-flow.yml'
   push:
     branches: [main]
@@ -31,16 +30,15 @@ on:
       - 'libs/ct-dap-client/**'
       - 'scripts/detect-siblings.sh'
       - 'justfile'
-      - '.github/sibling-pins'
       - '.github/workflows/mcr-dap-flow.yml'
   workflow_dispatch:
     inputs:
       native_recorder_ref:
-        description: 'codetracer-native-recorder ref (default: .github/sibling-pins)'
+        description: 'codetracer-native-recorder ref (default: dev)'
         required: false
         default: ''
       native_backend_ref:
-        description: 'codetracer-native-backend ref (default: .github/sibling-pins)'
+        description: 'codetracer-native-backend ref (default: dev)'
         required: false
         default: ''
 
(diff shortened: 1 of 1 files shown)


zah:

|---|---|---|

Neither is caused by this PR:

zah:

```
```

```
##[group]Run case "nix" in
^[[36;1mcase "nix" in^[[0m
^[[36;1m  *)^[[0m
^[[36;1m    exit 1^[[0m
^[[36;1m    ;;^[[0m
^[[36;1mesac^[[0m
```

Three things settle it:

```
    Updating crates.io index
```

---

@zah — flagging this here because I had published the wrong attribution on #636/#637 and your reading of `visual-replay-regression-gate` is the correct one. Independently confirmed from the raw log rather than from the rendered summary.

zah:

**Merge-base:** `af8c93c1c` (not `dev` tip); baseline is the dev run at that exact commit, [31975863960](https://github.com/metacraft-labs/codetracer/actions/runs/31975863960).

**Per-check evidence for `Codetracer CI`.** PR run [31976411554](https://github.com/metacraft-labs/codetracer/actions/runs/31976411554) is a job-for-job, failing-step-for-failing-step match to the baseline — same 18 failures at the same step names, same 5 cancellations, same 6 successes. The single deviation is `test-ui-tests (nixos)`, which failed *later* on the PR (step 16, `Run TypeScript Playwright UI tests`) than on this baseline (step 7, `Build CodeTracer (nix)`); step 16 is also where it fails on the adjacent dev baseline `521624086`, so that is flake variance in the PR's favour, not a regression.

**`mcr-dap-flow` — the check this PR exists to fix.** It is red on both sides, but for entirely different reasons, and the PR side is strictly further along.

Baseline (`MCR DAP Flow` on `main` @ `4a63621c9`, job 95693634178 — this workflow never fires on `dev`, since its triggers are `pull_request` and `push: branches: [main]`) dies at step 4 `Resolve sibling refs`, ~2.5 minutes in, before cloning a single sibling or installing Nix:

```
PINS=".github/sibling-pins"
  RECORDER_REF="$(grep '^codetracer-native-recorder ' "$PINS" | cut -d' ' -f2)"
grep: .github/sibling-pins: No such file or directory
##[error]Process completed with exit code 2.
```

Confirmed identical on a second contemporaneous baseline (job 95237474679).

The PR side (job 95236486294): `grep -n "sibling-pins"` over the whole 1497-line log returns **nothing**. Sibling resolution now succeeds:

```
metacraft-labs/codetracer-native-recorder -> dev (override)
metacraft-labs/codetracer-native-backend  -> dev (override)
metacraft-labs/codetracer-trace-format    -> dev (override)
metacraft-labs/codetracer-trace-format-nim -> dev (override)
Cloned 4 sibling(s) adjacent to the host checkout.
```

It then installs Nix, enters the dev shell, and runs step 6 for ~55 minutes before failing on something new and unrelated:

```
error: Failed to open archive (Source threw exception: error: unable to download
'https://api.github.com/repos/NixOS/nixpkgs/tarball/addf7cf5...': HTTP error 401
"message": "Bad credentials"
error: Recipe `ensure-ct-mcr` failed with exit code 1
```

That 401 is almost certainly token lifetime, not sibling resolution: `setup-nix` writes `access-tokens = github.com=$GH_TOKEN` using the App installation token minted at 22:30:21, and the nixpkgs fetch happens at 23:37:32 — ~67 minutes later, past the 60-minute expiry. It is reachable only *because* the job now runs for an hour instead of dying in two minutes.

**Confirmation the premise was true, not assumed:** `.github/sibling-pins` is absent from `git ls-tree -r af8c93c1c`; it was deleted by `e0c3a1593` ("resolve cross-repo sibling revisions from the workspace lock") eight weeks ago, and `mcr-dap-flow.yml` was not among the sites that commit migrated. It was the last consumer in the tree.

**The replacement is the sanctioned mechanism, verified against sibling call sites**, not asserted: `setup-dev-env@main` + `siblings:` with explicit `=dev` refs, as in `beam-flow.yml:92-123`, `cross-repo-tests.yml:66-76` and `:253-256`, and ten more `siblings:` blocks in `codetracer.yml`. The explicit-`=dev` convention (rather than bare names) is documented repo policy at `codetracer.yml:15-45`.

**Badges flipping green -> red:** none. `mcr-dap-flow` was already red on every run.

**Stated limitation:** because no `mcr-dap-flow` run has *ever* got past the pins step, there is no historical evidence about what else may be latent behind it. The nixpkgs 401 is the first thing this change exposed; there may be more. That is the expected shape of unblocking a gate that has been failing at step 4 for eight weeks.

PR #634, opened 2026-08-16
fix(test): let the JS ViewModel lane observe the failures it reports

zah:
`test-vm-js` reported results it could not actually see. Three defects
stacked, each hiding the next:

1. Nim auto-defines `nodejs` only for `nim js -r`. This lane compiles and
   runs as separate steps, so the define was absent, so
   `std/exitprocs.setProgramResult` was undeclared, so `std/unittest`
   substituted a no-op and a FAILING suite exited 0. Measured on this
   repo's Nim with a suite whose only test fails: `nim js` gives node
   exit 0, `nim js -d:nodejs` gives exit 1.

2. The compile redirected everything to `/dev/null`, so the compiler
   warning that names defect 1 outright -- "unittest will not give
   failing exit code on test failure" -- was discarded on every run, and
   was unrecoverable from `test-logs/test-vm-js.log`. On failure the
   recipe then recompiled and printed only lines matching `Error:`. This
   is the same reporting defect the native lane's own comment records
   having fixed, where filtering threw away the one line naming a missing
   library.

3. `output=$(node ...)` followed by `exitcode=$?` cannot observe a
   failure under this recipe's `set -e`: a non-zero `node` kills the loop
   before the assignment is read. So the `$exitcode` arm was dead even
   where defect 1 did not apply.

The lane still went red when a test printed `[FAILED]`, which is why this
survived: the human console report was doing all the work, while the
structured signal the recipe believed it was also checking did not exist.
A suite that crashed before printing anything, or that ran zero tests,
was scored `OK (0 tests)` -- the one result a test lane must never
invent. The native lane has had a `DID NOT RUN` guard for that case for a
while; this lane had none.

All four are fixed by bringing the lane to parity with the native one and
adding the define: compile with `-d:nodejs`, keep the compile output
instead of recompiling, capture node's status with `|| exitcode=$?`, and
report a run that produced no test results as a failure.

`ci/test/vm-js-lane-test.sh` pins it. Its shape contracts are static
checks over the justfile, so `ci-verdict` runs them from a stock runner
where they cannot be taken down by the outages that stop the lane itself
from executing. Its dynamic contracts compile the same deliberately
failing suite both ways and assert the exit codes differ, so the suite
carries its own mutation: if `-d:nodejs` ever stops being load-bearing,
the contract asserting it fails rather than quietly becoming cargo cult.
Where nim and node are absent those two are counted as skipped, never as
passed.

Each of the five contracts was verified to fail against the unfixed
recipe. Two needed the checks tightened first: the recipe's comments
necessarily quote the tokens under test, so a raw grep matched the
explanation rather than the command, and the `nim js` invocation spans
backslash continuations, so a line-oriented grep could not see the
redirect it was looking for. The suite now strips comments and folds
continuations, and guards against either transform emptying its own
haystack.

### DIFF
--- a/.github/workflows/codetracer.yml
+++ b/.github/workflows/codetracer.yml
@@ -3536,6 +3536,23 @@ jobs:
         # runner, regardless of the state of the heavy lanes.
         run: bash ci/test/visual-replay-private-cargo-preflight-test.sh
 
+      - name: Assert the JS ViewModel lane can observe its own failures
+        # `test-vm-js` believed it was checking node's exit code. It was not:
+        # without `-d:nodejs` std/unittest cannot set a failing exit code on
+        # the js backend, so a failing suite exited 0 -- and the compile
+        # redirected the compiler warning that says exactly that to
+        # /dev/null. The lane went red only when a test printed `[FAILED]`,
+        # so a suite that ran zero tests, or died before printing anything,
+        # scored `OK`.
+        #
+        # The shape half of this suite is a static check over the justfile, so
+        # it belongs here on a stock runner: the lane it guards runs inside the
+        # dev shell on a self-hosted runner, and is therefore exactly the kind
+        # of lane that can stop reporting without anyone noticing. The suite's
+        # toolchain-dependent half skips cleanly where nim and node are absent
+        # and is exercised in full by `viewmodel-tests`.
+        run: bash ci/test/vm-js-lane-test.sh
+
       - name: Assert the required jobs actually ran
         env:
           NEEDS_JSON: ${{ toJSON(needs) }}
--- a/ci/test/vm-js-lane-test.sh
+++ b/ci/test/vm-js-lane-test.sh
@@ -1,0 +1,221 @@
+#!/usr/bin/env bash
+#
+# vm-js-lane-test.sh — contract suite for the `test-vm-js` recipe in justfile.
+#
+# WHY THIS EXISTS
+# ---------------
+# The JS ViewModel lane reported test results it could not actually observe.
+# Three defects stacked, and each one hid the next:
+#
+#   1. `nim` auto-defines `nodejs` only for `nim js -r`. This lane compiles and
+#      runs as separate steps, so the define was absent, so
+#      `std/exitprocs.setProgramResult` was undeclared, so `std/unittest`
+#      substituted a no-op and a FAILING suite exited 0.
+#   2. The compile redirected everything to /dev/null, so the compiler's own
+#      "unittest will not give failing exit code on test failure" warning --
+#      which names defect 1 outright -- was discarded on every run.
+#   3. `output=$(node ...)` followed by `exitcode=$?` cannot observe a failure
+#      under the recipe's `set -e`: a non-zero `node` kills the loop before
+#      the assignment is read. So the `$exitcode` arm was dead even where
+#      defect 1 did not apply.
+#
+# The lane still went red when a test printed `[FAILED]`, which is why this
+# survived: the human console report was doing all the work, and the structured
+# signal the recipe believed it was also checking did not exist. A suite that
+# crashed before printing anything, or that ran zero tests, scored `OK`.
+#
+# The static contracts below pin the recipe's shape; the dynamic ones prove the
+# claim about `-d:nodejs` against the real toolchain rather than asserting it.
+#
+# DESIGN CONSTRAINT
+# -----------------
+# The static half is pure bash over the justfile, so it runs on a stock runner
+# with no Nim, no node and no dev shell -- the same rule the other CI gates
+# follow, and what lets `ci-verdict` run it. The dynamic half needs `nim` and
+# `node`; when they are absent it says so and is counted as skipped rather than
+# quietly passing.
+#
+# Run directly:  bash ci/test/vm-js-lane-test.sh
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_DIR}/../.." && pwd)"
+JUSTFILE="${REPO_ROOT}/justfile"
+
+pass_count=0
+skip_count=0
+
+fail() {
+	echo "vm-js-lane-test: FAIL: $1" >&2
+	shift
+	for line in "$@"; do echo "    ${line}" >&2; done
+	exit 1
+}
+
+ok() {
+	pass_count=$((pass_count + 1))
+	echo "  ok — $1"
+}
+
+skip() {
+	skip_count=$((skip_count + 1))
+	echo "  -- skipped: $1"
+}
+
+[ -f "${JUSTFILE}" ] || fail "justfile not found at ${JUSTFILE}"
+
+# Extract the `test-vm-js` recipe body: from its target line to the next
+# top-level target or comment block at column 0.
+recipe="$(awk '
+	/^test-vm-js:/ { inrec = 1; next }
+	inrec && /^[^[:space:]#]/ { exit }
+	inrec { print }
+' "${JUSTFILE}")"
+
+[ -n "${recipe}" ] || fail "could not extract the test-vm-js recipe from ${JUSTFILE}" \
+	"either the recipe was renamed or this extractor no longer matches it," \
+	"and every contract below would be vacuous"
+
+# The shape contracts below must read the recipe's CODE, never its prose.
+# This recipe carries a long comment explaining the very defects being pinned,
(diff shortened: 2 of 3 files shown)


zah:

|---|---|---|

Neither is caused by this PR:

zah:

**Why not wait:** `ci-verdict` publishes ~24h after a push and the fleet is capacity-capped. As I write this, the newest `dev` run sits at 22 queued / 0 running. Waiting produces no signal.

**CI gives essentially nothing here, and I want to be explicit.** On PR run [31976831750](https://github.com/metacraft-labs/codetracer/actions/runs/31976831750) *twenty* jobs — including all four lints, `CLAAssistant`, and crucially `ViewModel headless tests (native + JS)`, the lane this PR is about — failed at step 1 `Set up job`. That is GitHub `codeload` 429/503 while resolving action tarballs, before any workflow step ran:

```
##[error]Response status code does not indicate success: 429 (Too Many Requests).
Failed to download archive 'https://codeload.github.com/actions/create-github-app-token/tar.gz/fee1f7d...' after 3 attempts.
```

The baseline's `ViewModel headless tests` failed at step 7, i.e. inside the lane. Comparing "died at step 1" against "died at step 7" is a comparison of two absences, so I verified locally instead.

**Local verification #1 — the root cause, measured rather than argued.** The PR ships its own dynamic contract that compiles a deliberately-failing suite *both ways* and compares. Run against this repo's actual toolchain (`codetracer-nim/bin/nim`, node 22.22.2):

```
test-vm-js lane contracts
  ok — the recipe's nim js invocation carries -d:nodejs
  ok — node's exit status is captured with an || guard
  ok — a build that produces no test results is reported, not scored OK
  ok — the compile does not discard its diagnostics
  ok — the recipe still fails a suite on a non-zero exit code
  ok — with -d:nodejs a failing suite exits non-zero (got 1)
  ok — without the define that same suite exits 0 — the flag is load-bearing
  ok — compiling with -d:nodejs raises no setProgramResult warning

assertions: 8  skipped: 0  fail: 0
```

Lines 6 and 7 are the claim, confirmed by direct measurement: **the same failing suite exits 1 with `-d:nodejs` and 0 without it.** So yes — a failing JS suite was exiting 0, and `setProgramResult` was a no-op.

**Local verification #2 — the suite is not vacuous.** I copied the new `ci/test/vm-js-lane-test.sh` onto the *unfixed* `dev` tip and ran it there. It fails, on the right contract, with the right message:

```
vm-js-lane-test: FAIL: the recipe's nim js invocation carries -d:nodejs
    Without it std/unittest cannot set a failing exit code on the js
    backend, so node exits 0 for a failing suite and the exit-code arm
    below becomes dead code.
EXIT=1
```

Confirmed independently that `dev`'s `test-vm-js` recipe carries no `-d:nodejs`.

**On the third defect, which I think is the most easily missed:** `output=$(node ...)` followed by `exitcode=$?` genuinely cannot observe a failure under the recipe's `set -e` — the non-zero `node` kills the loop before the assignment is read. So the `$exitcode` arm was dead *twice over*, and `|| exitcode=$?` is the part that makes it readable at all. Fixing only the define would not have been enough.

**Two details in this suite that I want to call out as good practice**, because both are the exact failure mode this campaign keeps finding:
- It strips comments before grepping, because the recipe's own explanatory comment quotes `-d:nodejs` and `>/dev/null` — a raw grep passed against a deliberately re-broken recipe.
- It folds backslash continuations before grepping, because the `>/dev/null` sat on the last of five continuation lines and a per-line grep matched nothing, passing against the very defect it exists to catch.
- And it guards both transforms with a `grep -q 'nim js'` non-empty check so a broken stripper cannot make every contract pass vacuously.

**Badges flipping green -> red:** none *by conclusion* — `ViewModel headless tests (native + JS)` is already red on `dev`. But the honest expectation is that this lane will now report **more** failures than before, because it can finally see them, and a suite that runs zero tests or dies before printing will now say `DID NOT RUN` instead of `OK (0 tests)`. That is the entire point of the change; nothing that was truly passing starts failing.

Rebases cleanly onto current `dev`; the new `ci-verdict` step is pure bash whose dynamic half skips cleanly on a stock runner with no nim/node.

zah:
Post-merge follow-up: a defect in this PR that I missed before merging, plus corroboration of the root cause from the compiler source.

## Defect: the dynamic contracts never run in CI, and the header says otherwise

`ci/test/vm-js-lane-test.sh` states:

> *The suite's toolchain-dependent half skips cleanly where nim and node are absent and is exercised in full by `viewmodel-tests`.*

**That second clause is false.** On `dev` there is exactly one call site:

```
$ git grep -n vm-js-lane-test metacraft-labs/dev -- .github ci justfile
.github/workflows/codetracer.yml:3706:        run: bash ci/test/vm-js-lane-test.sh
```

— inside `ci-verdict`, whose `runs-on: ubuntu-latest` has no `nim`. Nothing in any workflow or justfile recipe invokes this script inside the dev shell, so the guard

```bash
if ! command -v nim >/dev/null 2>&1 || ! command -v node >/dev/null 2>&1; then
	skip "nim and/or node not on PATH; cannot verify the -d:nodejs claim here"
```

takes the skip branch on **every** CI run. The two contracts that actually prove `-d:nodejs` is load-bearing — the ones that make this suite more than a grep — are dead in CI.

This is uncomfortably close to the defect class this PR exists to fix: a check whose most valuable half cannot observe what it claims to check. The difference, and the reason I am not reverting, is that it is **not vacuous**. The five static contracts do run on `ubuntu-latest`, and I verified they have teeth: pointed at the pre-merge `justfile`, the suite fails on contract 1 with the correct message and exits 1. The regression this guards against is genuinely caught in CI. What is missing is the mutation check on the claim itself.

**Suggested fix (small):** add `bash ci/test/vm-js-lane-test.sh` to a job that runs inside the dev shell — `viewmodel-tests` is the natural home, and is what the comment already promises — or gate the skip so that being on a runner *with* nim is required somewhere. Either way the header comment should be corrected to match reality.

Two lesser notes from the same pass: contract 2's regex is pinned to the recipe's literal text (`node "\$cache/\$name\.js" 2>&1\) \|\| exitcode=`), so a benign reformat of the recipe breaks the gate rather than the contract; and the summary line prints a hardcoded `fail: 0`, which is harmless only because every failure path calls `exit 1`.

## Corroboration: the root cause, in the compiler source

The claim is exactly right, and it is three files deep. All in `codetracer-nim`:

`compiler/commands.nim:947-951` — one of only two auto-define sites, both tied to `-r`:
```nim
of "run", "r":
  processOnOffSwitchG(conf, {optRun}, arg, pass, info)
  if conf.backend == backendJs:
    # for now, -r uses nodejs, so define nodejs
    defineSymbol(conf.symbols, "nodejs")
```

`lib/std/exitprocs.nim:72` — the gate that makes `setProgramResult` exist at all:
```nim
when not defined(nimscript) and (not defined(js) or defined(nodejs)):
```

`lib/pure/unittest.nim:1236-1240` — the silent degradation:
```nim
when not declared(setProgramResult):
  {.warning: "setProgramResult not available on platform, unittest will not" &
    " give failing exit code on test failure".}
  template setProgramResult(a: int) =
    discard
```

So a failing suite prints `[FAILED]` and exits 0. Every call site (`unittest.nim:900, 1030, 1360`) becomes a no-op. Reproduced on the fork's nim 2.3.1 + node 22.22.2: without the define, node exits 0 and the compiler emits `unittest.nim(1237, 12) Warning: setProgramResult not available on platform`; with it, node exits 1 and there is no warning.

## What the lane will actually report now

Less alarming than I assumed when I merged. The dev baseline's step 7 ([run 31975936283](https://github.com/metacraft-labs/codetracer/actions/runs/31975936283), job 95235318997) shows **no failing tests at all** — `Native: 69 passed, 0 failed`, and on the JS side 54 `OK` plus a single compile error:

```
welcome_screen_vm_test.nim ... COMPILE ERROR
    .../nim-2.2.8/nim/lib/pure/osproc.nim(24, 8) Error: cannot export: quoteShell
JS: 54 passed, 1 failed
```

So there is no backlog of silently-failing JS suites about to surface. Worth noting CI uses stock nim 2.2.8 from the nix store here, not the `codetracer-nim` fork.

(One local run appeared to show an extra failure under the new recipe. Chased it down: `isonim_views_test.nim` fails identically with and without `-d:nodejs`; the apparent difference was a warm `/tmp/ct-nim-cache` from Aug 12. The failing case — `value-history button uses image-button markup and calls bridge`, `isonim_views_test.nim:1253-1254` — is real and pre-existing at dev tip, expecting `ct-button-image-sm-secondary` but getting `ct-button-image-xs-secondary`, and prints `[FAILED]`, so the old recipe flagged it too. Not caused by this PR.)

PR #635, opened 2026-08-16
fix(ci): scope the macOS ct-mcr credential instead of rewriting all of github.com

zah:
The macOS ct-mcr build installed a GLOBAL catch-all before invoking the
recorder's nix shell:

    git config --global url."https://x-access-token:$TOKEN@github.com/" \
      .insteadOf "https://github.com/"

That is the pattern the shared `setup-nix` stopped using, and it undid
that fix from inside the job. Unlike a per-checkout LOCAL header, global
config is inherited by `git clone`, and a rewritten URL carries the token
in argv, in the `.git/config` of every clone the rewrite touches, and in
git's own "unable to access 'https://x-access-token:...'" failure text.
It applied to every github.com fetch in the job -- nixpkgs, third-party
crates, any transitive git dependency -- none of which the token is for.
`setup-nix` evicts exactly this pattern from a reused `~/.gitconfig` at
the start of every job; these two lines put it straight back, after the
eviction had run.

They are also unnecessary, which is the part that had to be established
rather than assumed. `Setup dev env` already installs an owner-scoped
credential job-wide, and four links were checked end to end:

  * `Setup dev env` carries no `if:` on this job, so the macOS leg gets
    it as well as the nixos leg.
  * It runs the shared `setup-nix`, which exports
    `GIT_CONFIG_KEY_0=http.https://github.com/metacraft-labs/.extraHeader`
    through `$GITHUB_ENV`, making it job-wide rather than step-local.
  * The dependency being fetched is
    `github.com/metacraft-labs/lldb-sys.rs`. Git matches `http.<url>.*`
    by scheme, host, port and slash-delimited path prefix, so the
    `metacraft-labs/` scope covers it -- and covers nothing outside the
    org, which is the whole point of the narrower form.
  * `GIT_CONFIG_COUNT`/`KEY_n`/`VALUE_n` are ordinary environment
    variables, so they survive `nix develop --command` into cargo's child
    git processes. This was measured, not inferred: inside a
    `nix develop` shell with only those variables set, `git config
    --get-all` resolves the header from the environment alone.

`CARGO_NET_GIT_FETCH_WITH_CLI=true` is untouched -- it is what routes the
fetch through the git CLI at all, and cargo's default libgit2 transport
would ignore any gitconfig. The same scoped-header plus git-CLI
combination already fetches this exact dependency in
`visual-replay-regression-gate`, which additionally neutralises
`~/.gitconfig`, so the mechanism is in production use here today.

The two remaining catch-alls, in `push-tag` and `create-release`, are
deliberately left alone. They sit on the release path, which no PR or
`dev` run exercises, so the same reasoning cannot be demonstrated for
them the way it can here -- and an untested credential change on the
release path is the wrong trade. They are recorded for a maintainer to
check during a real release rather than changed blind.

### DIFF
--- a/.github/workflows/codetracer.yml
+++ b/.github/workflows/codetracer.yml
@@ -2497,12 +2497,44 @@ jobs:
         run: |
           set -euo pipefail
           cd "${{ github.workspace }}/../codetracer-native-recorder"
-          # Allow cargo to use git's credential helper so cloning the
-          # private ``lldb-sys.rs`` dependency picks up the App-token
-          # we set via ``GIT_CREDENTIALS`` below.
+          # Cargo's default libgit2 transport does not read gitconfig, so the
+          # git CLI has to be forced for the private ``lldb-sys.rs``
+          # dependency. The CREDENTIAL it then uses is already installed
+          # job-wide by ``Setup dev env`` above, so nothing is configured here.
+          #
+          # That step runs the shared ``setup-nix``, which exports
+          #
+          #   GIT_CONFIG_KEY_0=http.https://github.com/metacraft-labs/.extraHeader
+          #
+          # through $GITHUB_ENV. Four things make it sufficient here, and each
+          # was checked rather than assumed:
+          #
+          #   * ``Setup dev env`` is unconditional on this job -- it is not
+          #     gated on ``matrix.platform`` -- so the macOS leg gets it too.
+          #   * ``lldb-sys.rs`` lives at github.com/metacraft-labs/lldb-sys.rs,
+          #     and git matches ``http.<url>.*`` by scheme + host + port +
+          #     slash-delimited path prefix, so the metacraft-labs scope covers
+          #     it (and nothing outside the org).
+          #   * GIT_CONFIG_COUNT/KEY_n/VALUE_n are ordinary environment
+          #     variables, so they survive ``nix develop --command`` into
+          #     cargo's child git processes. Verified directly: inside a
+          #     ``nix develop`` shell, ``git config --get-all`` resolves the
+          #     header from the environment alone.
+          #   * The same scoped-header + git-CLI combination already fetches
+          #     this exact dependency in ``visual-replay-regression-gate``,
+          #     which builds its own per-repo header and neutralises
+          #     ~/.gitconfig entirely.
+          #
+          # What used to be here was a GLOBAL catch-all rewriting every
+          # github.com URL to carry the token as userinfo. Unlike a
+          # per-checkout local header, ``--global`` config IS inherited by
+          # ``git clone``, and it put the token into argv, into the
+          # ``.git/config`` of every clone it touched, and into git's own error
+          # messages -- for nixpkgs and every third-party fetch as well, none of
+          # which the token is for. ``setup-nix`` evicts exactly that pattern
+          # from a reused ~/.gitconfig at the start of the job; re-adding it
+          # here put it straight back.
           export CARGO_NET_GIT_FETCH_WITH_CLI=true
-          git config --global url."https://x-access-token:${{ steps.ci_token.outputs.token }}@github.com/".insteadOf "https://github.com/"
-          git config --global --add url."https://x-access-token:${{ steps.ci_token.outputs.token }}@github.com/".insteadOf "ssh://git@github.com/"
           nix develop --no-write-lock-file --command just build-ct-mcr
 
       - name: Build ct_gfx_player via codetracer-visual-replay nix develop (macOS)


zah:
Follow-up for a maintainer: the two remaining global catch-alls, and why they were NOT changed here.

`push-tag` and `create-release` each still run three of these before their real work:

```
git config --global url."https://x-access-token:$TOKEN@github.com/".insteadOf "git@github.com:"
git config --global url."https://x-access-token:$TOKEN@github.com/".insteadOf "ssh://git@github.com/"
git config --global url."https://x-access-token:$TOKEN@github.com/".insteadOf "https://github.com/"
```

They have the same three problems as the pair removed in this PR: `--global` config persists into whatever runs next on a runner that reuses its home directory, the token ends up in argv / `.git/config` / git error text, and the third line attaches it to every github.com fetch in the job rather than only the org’s own repositories.

**They are deliberately left in place.** Both jobs are on the release path (`push-tag` is tag-gated, `create-release` depends on it), so neither runs on a PR or on a `dev` push. Every argument that made the macOS removal safe is an argument I could *check* — that `Setup dev env` is unconditional on that job, that the dependency is inside `metacraft-labs/`, that `GIT_CONFIG_*` survives `nix develop`. None of them can be checked here without an actual release, and an unverifiable credential change on the release path is the wrong trade.

What a maintainer should check during the next real release, in order:

1. **Does the App installation hold `contents: write` on this repo?** This is the load-bearing unknown and it is independent of the catch-alls. The last tag (25.11.1) predates the switch to App tokens, so the tag push has never been exercised with the current credential. If `git push origin <tag>` returns 403, that is the App’s permission set — grant the installation `contents: write` rather than reintroducing a PAT.

2. **Confirm which credential actually authenticates the tag push.** It is the `http.https://github.com/.extraheader` that `actions/checkout` persists, not the userinfo these `insteadOf` lines inject. Verified at wire level against a local HTTP server with both configured: git sends the header, and on a 401 it retries with the *same* header and gives up — the userinfo never reaches the wire. That is why `push-tag` must keep `persist-credentials: true` (it does), and it is also why these three lines can most likely go.

3. **Only then remove them**, one job at a time, and re-run a release. If `create-release`’s `nix`/`wget`/`aws` steps need a github.com credential at all, prefer the owner-scoped `extraHeader` that `setup-nix` already exports job-wide over a global URL rewrite.

Nothing above blocks this PR: it changes neither job.

zah:

|---|---|---|

Neither is caused by this PR:

zah:

**Per-check evidence.** PR run [31976898853](https://github.com/metacraft-labs/codetracer/actions/runs/31976898853). Every red check is red on the baseline. PR-side-only reds are all pre-code infra with no parity value: `cross-process value-origin`, `ct-test cross-language provider gate`, `ct-test release gate`, `lint-nim`, `lint-nix`, `lint-rust`, `push-gpg-public-key`, `push-install-script`, `test-non-gui (nixos)`, `test-ui-tests (nixos)`, `test-ui-tests-rr`, `ViewModel headless tests` all died at step 1 `Set up job`, and `lint-bash` at step 4 `Install Nix` — GitHub `codeload` 429/503 while resolving action tarballs, before any workflow step ran.

The job this PR edits is the macOS leg of `test-ui-tests`, which is one of the step-1 casualties. So CI gives no evidence either way and I verified the mechanism directly instead.

**What I verified, against the actual action source rather than the description:**

1. `Setup dev env` on `test-ui-tests` is unconditional — no `if:`, not gated on `matrix.platform` — so the macOS leg does run it. Confirmed by reading the job.
2. It resolves to `metacraft-labs/metacraft-github-actions/setup-dev-env@main` with `env-flavor: nix`, which delegates to `setup-nix@main`, whose `Configure Git authentication` step runs `configure-git-auth.sh`.
3. That script does exactly what this PR relies on. From its own header: *"The credential is an `http.<url>.extraHeader`, so no URL anywhere carries it"* and *"It is exported through `GIT_CONFIG_COUNT`/`GIT_CONFIG_KEY_n`/`GIT_CONFIG_VALUE_n` via `$GITHUB_ENV`, which is job-scoped."* The header is scoped to `http.https://github.com/<owner>/.extraHeader` for `token-owner`, which defaults to `metacraft-labs` — and `lldb-sys.rs` lives at `github.com/metacraft-labs/lldb-sys.rs`, so it is covered.
4. The strongest corroboration: the same script *actively evicts* the pattern this PR removes. Lines 130-156 `--replace-all` and `--remove-section` the global `url."https://...@github.com/".insteadOf` entries out of a reused `~/.gitconfig`. The deleted lines were re-installing, one step later, precisely what `setup-nix` had just deleted. That is not a stylistic preference; the two were fighting.
5. `GIT_CONFIG_COUNT`/`KEY_n`/`VALUE_n` are ordinary environment variables, so they survive `nix develop --command` into cargo's child git processes. `CARGO_NET_GIT_FETCH_WITH_CLI=true` is retained, which is what makes cargo use the git CLI (and hence gitconfig/env) at all.
6. The same scoped-header pattern already fetches this exact dependency in `visual-replay-regression-gate` (`codetracer.yml` ~2816-2841), which builds its own per-repo `GIT_CONFIG_KEY_0` extraHeader and neutralises `~/.gitconfig` entirely.

**Why the old form was worse, concretely:** `--global` config *is* inherited by `git clone`, so the token went into argv, into the `.git/config` of every clone the job touched, and into git's error messages — for nixpkgs and every third-party fetch too, none of which the token is for.

**Residual risk, stated plainly:** the macOS `test-ui-tests` leg has not completed since this change, so the private `lldb-sys.rs` fetch has not been observed succeeding under the scoped header *in this job*. It has been observed succeeding under the same mechanism in `visual-replay-regression-gate`. If that leg comes back red at `Build ct-mcr`, this is the first place to look.

Issue #612, opened 2026-07-29
Omniscience "Add to Scratchpad" adds the value multiple times instead of once


pxor:
### Description

Clicking the `+` button on an Omniscience inline value (via the "Values:" popup
on a flow annotation) adds that value to the Scratchpad multiple times in a
single click. In the observed case, each single click adds the value **4 times**
instead of once: `i: 1 u32` appears 4 times (clicked once), `y: 2 Field` appears
4 times (clicked once), and `result: 5 Field` appears 8 times because it was
clicked twice — 4 duplicates per click, consistently.

### Expected behaviour (per spec)

Reference: `GUI/Core-Panes/Scratchpad-Pane.md` — "Adding Values"

> **Flow Values** — Right-click inline value → "Add to Scratchpad"

A single user action (click or right-click → menu item) on an Omniscience
inline value should add that value **exactly once** to the Scratchpad. The
Scratchpad entry should show the variable name, its value, and a close (X)
button:

```
result: 5 Field                                              [X]
```

If the same value is added again (e.g. across iterations), it should append
to the existing entry as a comma-separated list — not create a duplicate row.

### Current behaviour

A single click on the `+` button next to an Omniscience value adds it **4 times**.
The Scratchpad fills with identical rows:

```
[X] result: 5 Field
[X] result: 5 Field
[X] result: 5 Field
[X] result: 5 Field   ← all 4 from one click
```

The multiplier is consistently 4 per click regardless of which value is clicked.

### Root cause hypothesis

The click event handler on the `+` button is almost certainly being registered
multiple times — once per render/re-render of the Omniscience annotation or the
flow overlay. When the button is clicked, all N registered handlers fire,
producing N scratchpad entries. This is a classic event listener leak caused by
attaching handlers inside a render loop without removing the previous ones first.

The consistent multiplier of **4** likely corresponds to the number of times the
flow overlay re-rendered since the component mounted (e.g. once per step
navigation or once per loop iteration visible on screen).

### Steps to reproduce

1. Open a trace with a loop (e.g. the `looper` function in the Noir test program)
2. Navigate to a step inside the loop
3. Hover over a flow annotation to reveal the "Values:" popup with the `+` button
4. Click `+` on any value (e.g. `result: 5 Field`)
5. Check the Scratchpad — the value appears 4–8 times instead of once

### Spec reference

`GUI/Core-Panes/Scratchpad-Pane.md` — "Adding Values" table (Flow Values row),
[...]

Issue #555, opened 2026-06-16
Redraw issue on new file open


pxor:
**Issue:** Redraw Loop on File Open — Button Flickering

**Description:**
When opening a new file, the UI enters a continuous redraw loop causing buttons to blink and flicker (disappear/reappear). The issue is intermittent and does not occur consistently across all file opens.

**Steps to Reproduce:**
1. Open a new file
2. Observe buttons flickering/blinking during initial render
3. Note: Issue is non-deterministic and more likely with larger file redraws

**Expected Behavior:**
File should open and render smoothly without visual glitches or flickering UI elements.

**Actual Behavior:**
Buttons blink on and off rapidly due to excessive redraw cycles. The loop stops after a larger redraw operation completes.

**Environment Notes:**
- Intermittent issue (not reproducible on every file open)
- More pronounced with larger files/complex redraws
- Self-resolves after subsequent redraw operations

**Severity:** Medium — Impacts UX but self-corrects

https://github.com/user-attachments/assets/f5a16db4-7613-4eae-b66d-878388f99989

zah:
Fixed on `dev` in https://github.com/metacraft-labs/codetracer/commit/4b67003e2fe30008f12f752d173e68734a867966.

The caption toolbar was being torn down and rebuilt on every redraw — about
forty-five times per trace open. An earlier guard could never fire, because the
toolbar's host element is emitted by the menu shell and the shell clears its
container before each render, so the host was always freshly empty.

Leaving this open for you to verify.

PR #636, opened 2026-08-16
ci: stop the CLA check from booting a VM to do nothing

zah:
## What

The CLA workflow gated its only step with a **step-level** `if`. A step-level condition is evaluated *after* a runner has been assigned, so every comment on every issue in the repo still provisioned an ephemeral Linux VM, started the job, skipped the step, and tore the VM down.

This is not a hypothesis. Run [31920424241](https://github.com/metacraft-labs/codetracer/actions/runs/31920424241) occupied the ephemeral runner `garm-qjynbrvg9hfi` (label `eph-linux-x64`) and its job steps read:

| step | conclusion |
| --- | --- |
| Set up job | success |
| CLA Assistant | **skipped** |
| Complete job | success |

At the time of writing **25 such runs were queued simultaneously** on a runner class with **6 concurrent slots**. They came from 25 comments on 25 *different* bug-report issues — none of them pull requests, so none of them could ever have signed a CLA.

## Changes

1. **Move the condition to the job.** Job-level `if` is evaluated before scheduling, so no runner is requested at all. Also require `github.event.issue.pull_request`, so comments on plain issues no longer reach the job.
2. **`runs-on: ubuntu-latest`.** A CLA lookup is a checkout plus an API call — it uses nothing from the custom NixOS image. This repository is public, so hosted runners are free.
3. **Per-pull-request concurrency group**, keyed on the PR number and *not* `github.ref`. On `pull_request_target` and `issue_comment` the ref is the **base** branch, so a ref-keyed group would put every open PR in one group and let one PR's CLA run cancel another's. `cancel-in-progress: false` because the action commits the signature file back to `main`, and a cancelled commit can leave it half-written.

Also moves the deprecated `elixir-flow-dap.yml` redirect notice — two `echo` lines — off `eph-linux-x64` for the same reason.

## Effect

Frees the 25 queued `eph-linux-x64` jobs immediately and prevents the class of waste recurring. No CLA behaviour changes for actual pull requests.

## Validation

`actionlint` clean on both changed files.

### DIFF
--- a/.github/workflows/cla.yml
+++ b/.github/workflows/cla.yml
@@ -17,10 +17,43 @@ permissions:
 
 jobs:
   CLAAssistant:
-    runs-on: eph-linux-x64  # CIP-5: nix job → eph-linux-x64
+    # The gate is on the JOB, not on the step. It used to sit on the single
+    # step below, which meant every `issue_comment` on every *issue* in the
+    # repo still allocated a runner, booted it, skipped its only step, and
+    # tore it down. That is not hypothetical: run 31920424241 occupied the
+    # ephemeral runner `garm-qjynbrvg9hfi` to produce
+    # `CLA Assistant ... skipped`. With 25 such comments queued at once the
+    # workflow was holding the entire `eph-linux-x64` class (6 slots) hostage
+    # to do nothing. A job-level `if` is evaluated before scheduling, so no
+    # runner is requested at all.
+    #
+    # `github.event.issue.pull_request` is present only when the comment is on
+    # a pull request, so plain-issue chatter no longer reaches this job.
+    if: >-
+      github.event_name == 'pull_request_target' ||
+      (github.event.issue.pull_request != null &&
+       (github.event.comment.body == 'recheck' ||
+        github.event.comment.body == 'I have read the CLA Document and I hereby sign the CLA'))
+
+    # A CLA lookup is a git checkout plus an API call. It needs none of the
+    # custom NixOS image, so it has no business on the 6-slot self-hosted
+    # class. codetracer is a public repository, so `ubuntu-latest` is free.
+    runs-on: ubuntu-latest
+
+    # Grouped per pull request, NOT per `github.ref`: on `pull_request_target`
+    # and `issue_comment` the ref is the *base* branch, so a ref-keyed group
+    # would put every open PR in one group and let one PR's CLA run cancel
+    # another's.
+    #
+    # `cancel-in-progress: false` because this action commits the signature
+    # file back to the `main` branch; cancelling mid-commit can leave the
+    # signature store half-written. Queueing is the correct behaviour here.
+    concurrency:
+      group: cla-${{ github.event.pull_request.number || github.event.issue.number }}
+      cancel-in-progress: false
+
     steps:
       - name: "CLA Assistant"
-        if: (github.event.comment.body == 'recheck' || github.event.comment.body == 'I have read the CLA Document and I hereby sign the CLA') || github.event_name == 'pull_request_target'
         uses: contributor-assistant/github-action@v2.6.1
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/elixir-flow-dap.yml
+++ b/.github/workflows/elixir-flow-dap.yml
@@ -22,7 +22,10 @@ permissions:
 
 jobs:
   redirect-notice:
-    runs-on: eph-linux-x64  # CIP-5: nix job → eph-linux-x64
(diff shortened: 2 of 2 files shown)


zah:
### Coordination note — this fix only bites once it reaches `stable`

`issue_comment` and `pull_request_target` workflows always execute the copy of the workflow file on the repository's **default branch**. codetracer's default branch is **`stable`**, not `dev`.

So merging this into `dev` changes nothing on its own — the CLA workflow will keep running the old `stable` copy and keep booting a VM per comment until `dev` is promoted to `stable`. `dev` is currently 61 ahead of / 0 behind `stable`, so a normal promotion carries it.

### Evidence the leak is ongoing, not historical

While this PR was open I measured and then cancelled **32** queued no-op `CLA Assistant` runs on this repo:

- **25** from a bulk of "Fixed on `dev` in …" comments posted on plain issues #555–#612 (none is a pull request, so none could ever sign a CLA);
- **7** from ordinary pull-request comments posted since — I checked all 7 bodies; none is `recheck` or the signature phrase.

Every one of them would have provisioned an `eph-linux-x64` VM, skipped its only step, and torn the VM down. A 33rd has already appeared since. The leak rate is roughly one wasted VM per comment on this repository.

### Pre-existing failures on this PR

`workspace-lock-freshness` and `visual-replay-regression-gate` fail here. Both fail identically on `dev` tip — same job, same failing step, same log line — across the last 5 `dev` pushes:

- `workspace-lock-freshness` → failing step `Assert this commit has a published workspace lock`
- `visual-replay-regression-gate` → failing step `Run visual replay regression gate`, log line `::error::env-flavor must be one of: nix, windows-diy, reprobuild (got 'nix')`

Neither is caused by this change.

### Overlap check

This PR touches only `cla.yml` and `elixir-flow-dap.yml`. It does **not** touch `codetracer.yml` credential steps, `mcr-dap-flow.yml`, or `ci/verdict/`, so it should not conflict with the credential-boundary work in flight.

zah:

```
```

```
##[group]Run case "nix" in
^[[36;1mcase "nix" in^[[0m
^[[36;1m  *)^[[0m
^[[36;1m    exit 1^[[0m
^[[36;1m    ;;^[[0m
^[[36;1mesac^[[0m
```

Three things settle it:

```
    Updating crates.io index
```

zah:

**Why not wait:** `ci-verdict` publishes ~24h after a push (it waits on jobs whose `runs-on` label has no registered runner) and the fleet is capacity-capped. Waiting produces no signal.

**Per-check evidence.** PR run [31976991694](https://github.com/metacraft-labs/codetracer/actions/runs/31976991694). Every red check is red on the baseline too. The PR-side-only reds are all pre-code infra and therefore give no parity evidence in either direction:

- `ct-test cross-language provider gate`, `push-gpg-public-key`, `test-non-gui (nixos)`, `test-ui-tests (nixos)`, `test-ui-tests-rr`, `ViewModel headless tests`, `lint-rust` — step 1 `Set up job`, GitHub `codeload` 429/503 resolving action tarballs before any step ran.
- `lint-nim` step 3 `Checkout`, `lint-nix` step 4 `Install Nix` — same cause.
- `CLAAssistant` on the PR run failed at step 1 `Set up job`, same 429. That is the check this PR changes, so it is a comparison of two absences and I did not rely on it.

**What I relied on instead (verified, not assumed):**

1. *The mechanism.* A step-level `if` is evaluated after a runner is assigned; a job-level `if` is evaluated before scheduling. That is the whole point of the change and it is GitHub-documented behaviour, corroborated by the cited run 31920424241 which occupied `garm-qjynbrvg9hfi` only to report the step skipped.
2. *`runs-on: ubuntu-latest` is actually available to this repo.* Not taken on faith: `ubuntu-latest` is already used by two jobs in `codetracer.yml` on `dev`, and one of them (`ci-verdict`) demonstrably gets a runner and executes real steps in every recent run — it fails at step 5/7, i.e. well past scheduling. The repo is public, so GitHub-hosted minutes are free.
3. *The gate is not narrower than it needs to be.* The job-level `if` keeps `github.event_name == 'pull_request_target'` as an unconditional pass, so the CLA check still runs on every pull request exactly as before. The added `github.event.issue.pull_request != null` conjunct only removes the case of a plain-issue comment matching the magic strings — which could never have produced a meaningful CLA result anyway.
4. Parsed the merged `cla.yml` after rebasing onto current `dev`: job-level `if` present, no step-level `if` remaining, `runs-on: ubuntu-latest`.

**Second change in this PR:** `elixir-flow-dap.yml`'s `redirect-notice` job (two `echo` lines, no nix content) also moves off `eph-linux-x64` to `ubuntu-latest`. Same reasoning, no behaviour change.

**Badges flipping green -> red:** none. `CLAAssistant` continues to run on every PR.

PR #637, opened 2026-08-16
ci: collapse superseded pull-request runs

zah:
## What

Only the newest push to a pull request is worth validating, but every intermediate push was holding its own set of runners until it finished. On a fleet capped at six concurrent `eph-linux-x64` slots that is capacity spent producing answers nobody will read.

Adds a concurrency group to the pull-request validation workflows:

```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
  cancel-in-progress: true
```

## The group key is deliberate

`github.head_ref` is populated **only** on `pull_request` and is empty everywhere else. So:

- **pull requests** collapse per source branch — a new push cancels the run it superseded;
- **pushes to `main`/`dev`/`stable`**, the **nightly cron**, and **manual dispatches** fall through to `github.run_id`, which is unique per run. Each keeps a group of its own and is **never** cancelled.

That exemption is the point. `dev` is the integration branch and each commit's build result is independently meaningful when bisecting, so collapsing mainline pushes would destroy history rather than save capacity. The same reasoning protects the nightly `request-panel-sidecar-retirement` run, which is a standing proof whose data points must not be dropped because a PR happened to be busy.

Using `github.ref` instead — as is the common cargo-culted form — would have been wrong here: on `push` the ref *is* the mainline branch, so all mainline pushes would share one group and cancel each other.

## Per-workflow judgement

| workflow | triggers | category | applied |
| --- | --- | --- | --- |
| `codetracer.yml` | PR + push main/dev/stable | PR validation (widest fan-out: Windows + Linux + macOS) | yes, PR-only |
| `cross-repo-tests.yml` | PR + push + dispatch | PR validation | yes, PR-only |
| `beam-flow.yml` | PR + push main + dispatch | PR validation | yes, PR-only |
| `request-panel-sidecar-retirement.yml` | PR + push + nightly cron + dispatch | PR validation, but nightly is a data series | yes, PR-only; cron exempt via `run_id` |
| `elixir-flow-dap.yml` | dispatch only | deprecated redirect; each dispatch meaningful | no |
| `cla.yml` | `pull_request_target` + `issue_comment` | commits a signature file | separate PR, per-PR group, `cancel-in-progress: false` |
| `mcr-dap-flow.yml` | — | owned by another change in flight | not touched |

No publishing, release, tagging, deploy or cache-push workflow received a cancelling group.

## Validation

`actionlint` output is **byte-identical** before and after on every changed file (the repo has pre-existing warnings about the unknown custom `eph-*` labels; this change adds none).

`codetracer.yml` is in its own commit so it can be dropped independently if it conflicts with the credential-step work in flight on that file.

### DIFF
--- a/.github/workflows/beam-flow.yml
+++ b/.github/workflows/beam-flow.yml
@@ -41,6 +41,17 @@ on:
         required: false
         default: ''
 
+# Collapse superseded pull-request runs only.
+#
+# `github.head_ref` is set on `pull_request` and empty everywhere else, so PR
+# runs group per source branch and a new push cancels the run it superseded.
+# Pushes to `main` and manual dispatches fall through to `github.run_id`,
+# which is unique per run: each gets its own group and is never cancelled, so
+# the per-commit history on the mainline stays complete.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 permissions:
   contents: read
   id-token: write
--- a/.github/workflows/codetracer.yml
+++ b/.github/workflows/codetracer.yml
@@ -7,6 +7,23 @@ on:
       - stable
   pull_request:
 
+# Collapse superseded pull-request runs only.
+#
+# This is the widest workflow in the repository -- Windows, Linux and macOS
+# legs across three runner classes -- so a superseded pull-request run is the
+# single most expensive thing the fleet can be doing.
+#
+# `github.head_ref` is set on `pull_request` and empty everywhere else, so PR
+# runs group per source branch and a new push cancels the run it superseded.
+# Pushes to `main`/`dev`/`stable` fall through to `github.run_id`, which is
+# unique per run: every mainline commit keeps a group of its own and is never
+# cancelled. That is deliberate -- `dev` is the integration branch and each
+# commit's build result is independently meaningful for bisecting, so
+# collapsing mainline pushes would destroy history rather than save capacity.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 permissions:
   contents: read
   id-token: write
--- a/.github/workflows/cross-repo-tests.yml
+++ b/.github/workflows/cross-repo-tests.yml
@@ -25,6 +25,17 @@ on:
         required: false
         default: ''
 
+# Collapse superseded pull-request runs only.
+#
+# `github.head_ref` is set on `pull_request` and empty everywhere else, so PR
+# runs group per source branch and a new push cancels the run it superseded.
+# Pushes to `main`/`dev`/`stable` and manual dispatches fall through to
+# `github.run_id`, which is unique per run: each gets its own group and is
+# never cancelled, so the per-commit history on the mainlines stays complete.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 permissions:
   contents: read
   id-token: write
--- a/.github/workflows/request-panel-sidecar-retirement.yml
+++ b/.github/workflows/request-panel-sidecar-retirement.yml
@@ -46,6 +46,19 @@ on:
     - cron: '0 3 * * *'
   workflow_dispatch:
 
+# Collapse superseded pull-request runs only.
+#
+# `github.head_ref` is set on `pull_request` and empty everywhere else, so PR
+# runs group per source branch and a new push cancels the run it superseded.
+# Pushes to `main`/`dev`/`stable`, the nightly cron and manual dispatches fall
+# through to `github.run_id`, which is unique per run: each gets its own group
+# and is never cancelled. That matters here -- this workflow is the standing
+# proof that no recorder writes a sidecar manifest, so a nightly result is a
+# data point that must not be discarded because a PR happened to be busy.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
 permissions:
   contents: read
   id-token: write


zah:
### Coordination note

`codetracer.yml` is in its **own commit** (`ci: collapse superseded runs of the main Codetracer CI workflow`) precisely so it can be dropped with `git revert` / `rebase --onto` if it collides with the credential-step work in flight on that file. This PR adds only a top-level `concurrency:` block between `on:` and `permissions:` — it does not touch any job or step, `mcr-dap-flow.yml`, or `ci/verdict/`.

### Not merging yet — the bar isn't met

Every functional check on this PR is still `QUEUED`. A queued check is not a passing check, so this is not ready to merge.

Two checks have completed, both **failing**, and both fail identically on `dev` tip across the last 5 `dev` pushes — same job, same failing step, same log line:

| check | failing step | same on `dev` tip |
| --- | --- | --- |
| `workspace-lock-freshness` | `Assert this commit has a published workspace lock` | yes |
| `visual-replay-regression-gate` | `Run visual replay regression gate` — `::error::env-flavor must be one of: nix, windows-diy, reprobuild (got 'nix')` | yes |

Neither is caused by this change.

### What blocks demonstration

The remaining checks include legs on `eph-macos-arm64` and `eph-linux-arm64`, runner classes that have served no jobs since Aug 1. Those legs cannot complete until the runner-class outage is resolved, so this PR cannot be shown green on the standard bar before then.

zah:

```
```

```
##[group]Run case "nix" in
^[[36;1mcase "nix" in^[[0m
^[[36;1m  *)^[[0m
^[[36;1m    exit 1^[[0m
^[[36;1m    ;;^[[0m
^[[36;1mesac^[[0m
```

Three things settle it:

```
    Updating crates.io index
```

zah:

**Why not wait:** `ci-verdict` publishes ~24h after a push because it waits on jobs whose `runs-on` label has no registered runner, and the fleet is capacity-capped (`dev` has not been green in 100+ runs). Waiting would not produce a signal.

**Merge-base:** `521624086`. Baseline for comparison is the dev run at that exact commit, [31975936283](https://github.com/metacraft-labs/codetracer/actions/runs/31975936283).

**Per-check evidence:** this PR adds only a top-level `concurrency:` block to four workflow files — no job, step, script or source file is touched (`git diff --stat`: 4 files, +52/-0). Every failing check on the PR run [31977007250](https://github.com/metacraft-labs/codetracer/actions/runs/31977007250) is present and failing on the baseline. The PR-side deltas are all pre-code infra:

- `lint-bash`, `lint-rust`, `ct-test cross-language provider gate`, `test-non-gui (nixos)`, `test-ui-tests (nixos)`, `test-ui-tests-rr`, `ViewModel headless tests` failed at step 1 `Set up job` — GitHub `codeload` 429/503 while resolving action tarballs, before any workflow step ran. Verbatim: `##[error]Response status code does not indicate success: 429 (Too Many Requests).` / `Failed to download archive 'https://codeload.github.com/actions/create-github-app-token/tar.gz/fee1f7d...' after 3 attempts.`
- `lint-nim`, `lint-nix` failed at step 3 `Checkout`, same cause.

Those give no parity evidence either way, so the change was verified by construction instead: `concurrency` is workflow-level GitHub configuration with no runtime behaviour to test. The one substantive question — does this collapse mainline history? — is answered by the expression itself: `github.head_ref` is set only on `pull_request`, so pushes to `main`/`dev`/`stable`, the nightly cron and `workflow_dispatch` all fall through to `github.run_id`, which is unique per run and therefore never groups with anything. Mainline per-commit results are preserved; only superseded PR runs collapse.

Verified locally that all four files still parse as valid YAML after rebasing onto current `dev` (`6c490ffaf`), and that the rebase is clean.

**Known residual:** for `pull_request` events the group key is the source branch name, so two forks pushing branches of the same name would share a group. Acceptable, and no worse than the GitHub-documented default pattern.

zah:
Post-merge note: this change is worth more than its description claims, and the reason is worth recording.

I merged this on the reasoning that collapsing superseded PR runs saves fleet capacity — true, but incomplete. While gathering per-check evidence across this campaign I found that the dominant failure mode in `codetracer` CI right now is **not** a self-hosted runner shortage. It is GitHub rejecting **action tarball downloads**:

```
##[warning]Failed to download action 'https://codeload.github.com/actions/create-github-app-token/tar.gz/fee1f7d...'.
  Error: Response status code does not indicate success: 429 (Too Many Requests).
##[warning]Back off 21.798 seconds before retry.
...
##[error]Response status code does not indicate success: 429 (Too Many Requests).
##[error]Failed to download archive ... after 3 attempts.
```

Also seen as `503 (Service Unavailable)` on `metacraft-labs/reprobuild` tarballs. This happens during **step 1, `Set up job`** — before any workflow step runs — and it accounted for the majority of red checks on every campaign PR I examined (twenty of them on one run alone).

**Why that makes this PR load-bearing.** `codetracer.yml` is the widest workflow in the repo: ~37 jobs, each resolving several actions, each resolution a separate `codeload` fetch with up to 3 retries. A superseded run that keeps going is not merely occupying runners — it is **generating the rate-limit pressure that makes the run replacing it fail at step 1**. Collapsing superseded PR runs cuts those fetches roughly in proportion to how often people push to a branch.

So this is not only a capacity optimisation; it is a partial mitigation for the single most common cause of red checks in this repo. I did not know that when I wrote the merge comment, and the rationale should survive in a place someone will find it.

**What it does not do:** mainline pushes are deliberately not collapsed (`github.head_ref` is empty there, so each falls through to a unique `github.run_id`), so pushes to `dev` still each generate a full set of fetches. That was the right call for bisectability, but it means the mitigation only covers the PR side.

**Not to be confused with a different, genuinely separate defect:** `Error: [@octokit/auth-app] appId option is required`. That one fails at a *named* `actions/create-github-app-token@v2` step with exit 127, not at `Set up job`, and its cause is a repo not being registered with the CI-token configurator. I searched 18 failing jobs across three `codetracer` runs and found **zero** occurrences of it — `codetracer` is not affected. See my note on `infra` #1307 for where it is real.

**The discriminator, for anyone triaging these later:** look at *which* step failed.
- `Set up job` -> action tarball download, `429`/`503`. Infrastructure throttling; this PR helps.
- a named mint step -> `appId` missing. Repo registration; `infra` #1307 is the fix.

They look similar in a check list and have nothing to do with each other.

PR #638, opened 2026-08-16
fix(ci): stop persisting the checkout credential in seven git-free jobs

zah:
`actions/checkout` defaults to `persist-credentials: true`, which leaves
the job's token in `<workspace>/.git/config` as an `http.extraHeader`
after the action finishes. On runners that reuse their work tree that
credential outlives the step that needed it, for no benefit in a job that
never talks to a git remote again.

Seven such jobs are opted out here. Each was traced step by step to the
end of the job, not sampled:

  push-gpg-public-key       gpg --armor --export, then aws s3 cp
  push-install-script       aws s3 cp
  windows-named-pipe-tests  rust-toolchain, then cargo test --bin simple-tui
  appimage-cross-distro     nix profile install, aws s3 cp, then
                            scripts/test-appimage-cross-distro.sh
  origin-dap-macos          setup-nix, setup-db-backend-siblings, then
  origin-dap-macos-nightly  just test-origin-dap
  ci-verdict                five bash self-test steps, all static checks

No later step in any of them runs a git command against a remote. The
scripts they invoke were read rather than assumed: neither
`scripts/test-appimage-cross-distro.sh` nor `scripts/test-origin-dap.sh`
contains a git invocation at all, and `src/tui/Cargo.toml` declares no
git dependencies, so the `cargo test` fetches nothing over git.

Two things make this narrower than it looks, and both are why the rest of
the workflow is left alone for now:

  * The persisted header is written to the CHECKOUT's `.git/config`, so
    it only ever applied to a git process whose repository is this
    checkout. `clone-repo` / `clone-siblings` build their own step-scoped
    credential, and `git clone` does not inherit an enclosing repo's
    local config, so neither ever depended on it.
  * `nix develop` is unaffected in either direction. Nix's internal
    fetcher does not read gitconfig; its credentials come from the netrc
    and access-tokens that `setup-nix` installs. So the jobs above cannot
    lose a Nix fetch by dropping this.

`push-tag` deliberately keeps its credentials: its "Create tag" step runs
`git push origin` with the checkout as its working directory, against a
plain github.com remote, and the persisted header is the credential that
authenticates it. That is the one place in these workflows where a git
remote is contacted from inside the checkout, and it is the reason this
change is applied job by job rather than swept across the file -- a wrong
`persist-credentials: false` does not fail at lint time or at checkout
time, only later, at push time.

Jobs whose later steps call large toolchain bootstraps that could not be
enumerated exhaustively -- the `windows-diy` lane and the non-nix macOS
legs, which also have no `setup-nix` credential to fall back on -- are
left at the default rather than guessed at.

### DIFF
--- a/.github/workflows/codetracer.yml
+++ b/.github/workflows/codetracer.yml
@@ -706,6 +706,14 @@ jobs:
 
       - name: Checkout
         uses: actions/checkout@v5
+        with:
+          # No later step in this job talks to a git remote, so the token has
+          # nothing to do after the checkout and should not outlive it: these
+          # runners reuse their work tree, and actions/checkout otherwise
+          # leaves the credential in `.git/config` for whatever runs next.
+          # (`nix develop` is unaffected either way -- Nix's fetcher does not
+          # read gitconfig; its credentials come from netrc/access-tokens.)
+          persist-credentials: false
 
       - name: Install Rust toolchain
         uses: dtolnay/rust-toolchain@stable
@@ -1372,6 +1380,13 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v5
         with:
+          # No later step in this job talks to a git remote, so the token has
+          # nothing to do after the checkout and should not outlive it: these
+          # runners reuse their work tree, and actions/checkout otherwise
+          # leaves the credential in `.git/config` for whatever runs next.
+          # (`nix develop` is unaffected either way -- Nix's fetcher does not
+          # read gitconfig; its credentials come from netrc/access-tokens.)
+          persist-credentials: false
           submodules: recursive
           token: ${{ steps.ci_token.outputs.token }}
 
@@ -1419,6 +1434,13 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v5
         with:
+          # No later step in this job talks to a git remote, so the token has
+          # nothing to do after the checkout and should not outlive it: these
+          # runners reuse their work tree, and actions/checkout otherwise
+          # leaves the credential in `.git/config` for whatever runs next.
+          # (`nix develop` is unaffected either way -- Nix's fetcher does not
+          # read gitconfig; its credentials come from netrc/access-tokens.)
+          persist-credentials: false
           submodules: recursive
           token: ${{ steps.ci_token.outputs.token }}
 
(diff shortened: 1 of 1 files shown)


zah:

|---|---|---|

Neither is caused by this PR:

zah:

**Per-check evidence.** PR run [31977078531](https://github.com/metacraft-labs/codetracer/actions/runs/31977078531). Every red check is red on the baseline. PR-side-only reds are all pre-code infra and give no parity evidence:

- `cross-process value-origin`, `ct-test cross-language provider gate`, `lint-rust`, `push-gpg-public-key`, `push-install-script`, `test-non-gui (nixos)`, `test-ui-tests (nixos)`, `test-ui-tests-rr` — step 1 `Set up job`, GitHub `codeload` 429/503 resolving action tarballs before any step ran.
- `ct-test release gate` step 3 `Checkout`, `lint-bash` step 4 `Install Nix` — same cause.

Notably `push-gpg-public-key` and `push-install-script`, two of the seven jobs this PR edits, never reached a step. So CI does not tell us this change is safe, and I verified it by inspection instead.

**What I verified locally.**

1. *Exactly seven jobs gained the flag, and they are the seven claimed.* Diffed the `persist-credentials: false` site set between the merge-base and the PR head: added = `appimage-cross-distro`, `ci-verdict`, `origin-dap-macos`, `origin-dap-macos-nightly`, `push-gpg-public-key`, `push-install-script`, `windows-named-pipe-tests`. Eight other jobs already had it; none were removed.

2. *The load-bearing claim — "no later step in this job talks to a git remote" — holds for all seven.* Walked every step of each job:
   - `push-gpg-public-key`: `Import GPG key for signing commits` (configures signing only, no remote) then `Upload public key`, which is `gpg --armor --export` piped to `aws s3 cp` against the R2 bucket. No git remote.
   - `push-install-script`: `Upload script` is `aws s3 cp install-on-distributions.sh s3://...`. Despite the job names, neither "push" is a `git push`.
   - `origin-dap-macos` / `origin-dap-macos-nightly`: the only remote access is `Setup db-backend siblings`, a composite action that receives `gh-token: ${{ steps.ci_token.outputs.token }}` explicitly and does not depend on the outer checkout's persisted credential.
   - `windows-named-pipe-tests`: cargo test only. `appimage-cross-distro`: AWS CLI + docker. `ci-verdict`: five local assertion scripts.
   - No `git push`/`fetch`/`pull`/`clone`/`ls-remote`/`submodule` invocation appears in any later step of any of the seven.

3. `submodules: recursive` on two of them is unaffected: `actions/checkout` fetches submodules inside the checkout step, while the credential is still live.

4. Workflow YAML parses after rebasing onto current `dev` (`6c490ffaf`); rebase is clean.

**Why this matters:** these runners reuse their work tree, so a persisted token outlives the job that minted it and is readable by whatever runs next on that machine.

PR #639, opened 2026-08-16
fix(ci): record the non-nix build lane as retired instead of failing on it

zah:
`sibling-provisioning-test.sh` asserted that the workflows still contain
at least one `./non-nix-build/build.sh` step. That assertion existed to
stop its own scanner going quietly blind: the contract underneath it
pairs every non-nix build with a `codetracer-trace-format` checkout, and
a scanner that matches nothing would report that pairing as holding
vacuously.

The lane is now genuinely gone. 6fee21a17 moved the macOS non-gui and UI
legs into the nix dev shell on eph-macos-arm64, and the five remaining
mentions of `./non-nix-build/build.sh` in the workflows are all prose in
comments recording that migration -- there is no call site left to pair.
So the guard fired on every run, and since `ci-verdict` runs this suite,
it made that job red for a reason unrelated to anything it gates.

Replacing `> 0` with an exact expected count keeps the guard honest in
both directions rather than deleting it. Bringing a non-nix leg back now
fails this contract until the count is updated -- which is exactly the
moment someone has to confirm the pairing contract still holds, and that
contract is the one that originally caught a macOS leg building with no
sibling checkout and dying on "codetracer-trace-format checkout not
found". Verified by injecting a synthetic non-nix step: both the count
contract and the pairing contract fail, as intended.

### DIFF
--- a/ci/test/sibling-provisioning-test.sh
+++ b/ci/test/sibling-provisioning-test.sh
@@ -437,12 +437,35 @@ for wf in "${SIBLING_SOURCE_FILES[@]}"; do
 	done <"$wf"
 done
 
-if [ "$build_sites" -gt 0 ]; then
-	ok "the workflows still run the non-nix build ($build_sites call sites)"
+# How many `./non-nix-build/build.sh` call sites this workflow is EXPECTED to
+# have. The Homebrew-native macOS legs were migrated into the nix dev shell on
+# eph-macos-arm64 (6fee21a17), so the answer is now zero and the pairing
+# contract below has nothing left to pair.
+#
+# This is asserted as an exact count rather than dropped, because the check it
+# guards -- "the scanner still matches the thing it scans for" -- is still
+# worth having in both directions. Requiring `> 0` reported the retirement as a
+# failure on every run. Deleting the check outright would mean a re-introduced
+# non-nix leg silently reinstates the original defect (a build.sh with no
+# codetracer-trace-format sibling, which fails with "checkout not found"). An
+# exact count keeps that loud: bring the lane back and this fails until the
+# number is updated, which is precisely the moment someone must confirm the
+# pairing contract below still holds.
+EXPECTED_BUILD_SITES=0
+
+if [ "$build_sites" -eq "$EXPECTED_BUILD_SITES" ]; then
+	if [ "$build_sites" -eq 0 ]; then
+		ok "the non-nix build lane is retired, as recorded (0 call sites)"
+	else
+		ok "the workflows still run the non-nix build ($build_sites call sites)"
+	fi
 else
-	fail "the workflows still run the non-nix build" \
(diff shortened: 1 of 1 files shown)


zah:

|---|---|---|

Neither is caused by this PR:

zah:

**Primary evidence is local, and it is direct.** This PR changes exactly one file, `ci/test/sibling-provisioning-test.sh`, which is precisely what `ci-verdict` step 5 (`Assert every job's sibling checkout is self-sufficient`) executes. I ran it against the *current* `dev` tip (`6c490ffaf`), not just the merge-base, because `dev` has since grown that script by +88 lines:

```
=== dev tip, unmerged ===
  FAIL the workflows still run the non-nix build
         no ./'non-nix-build/build.sh' step was found -- either the macOS lane was
1 of 18 assertions failed          (exit 1)

=== dev tip + this PR ===
  ok   the non-nix build lane is retired, as recorded (0 call sites)
  ok   every non-nix build leg provisions codetracer-trace-format first
all 18 assertions passed           (exit 0)
```

Same result at the merge-base itself (1 of 16 failed -> 16 of 16 passed). So the claim in the description is confirmed end to end: this assertion was red on *every* run since `6fee21a17` retired the lane, and it is the only one of the 18 that was failing.

**Corroborating CI evidence.** On the PR run [31977177344](https://github.com/metacraft-labs/codetracer/actions/runs/31977177344), `ci-verdict` no longer fails at step 5 — it now gets past it and fails at step 7, `Assert the required jobs actually ran`. On the baseline it fails at step 5. That is the change taking effect in production. `ci-verdict` remains red on both sides; step 7 fails because the starved fleet never ran the required jobs, which is the pre-existing capacity outage and not something this PR claims to fix.

**Other PR-side reds, all pre-code infra, none giving parity evidence:**
- `lint-rust`, `ct-test cross-language provider gate`, `push-gpg-public-key`, `test-non-gui (nixos)` — step 1 `Set up job`, `codeload` 429/503 while resolving action tarballs.
- `lint-bash`, `lint-nim` step 4 `Install Nix`; `lint-nix`, `cross-process value-origin` step 3 `Checkout` — same cause.
- `reprobuild-linux-smoke` is green on the baseline and red here. It is **not** this PR: it died fetching an unrelated third-party crate, `remote: Internal Server Error` / `unable to access 'https://github.com/mystenlabs/anemo.git/': The requested URL returned error: 500`. This PR touches one bash file that `reprobuild-linux-smoke` does not run.

**Design note I checked rather than assumed:** the change asserts an *exact* count (`EXPECTED_BUILD_SITES=0`) rather than deleting the check. That keeps the scanner honest in both directions — reintroducing a non-nix leg fails loudly until someone confirms the `codetracer-trace-format` pairing contract still holds. Deleting it would have made the reintroduction silent. Good call.

PR #640, opened 2026-08-16
fix(ci): judge native-backend's lock by what it pins, not by --locked

zah:
## The defect

`visual-replay-regression-gate` is red on `dev` and on every open PR. Everything
upstream of the failure passes (Playwright 19/19 fake + 1/1 real, Nim lifecycle
3/3, adversarial contracts, the private-Cargo preflight's own auth contract);
the whole failure is one command in
`ci/test/visual-replay-private-cargo-preflight.sh`:

```
cargo fetch --locked --manifest-path <native-backend>/Cargo.toml
error: cannot update the lock file ... because --locked was passed to prevent this
```

`codetracer-native-backend/Cargo.toml` routes one dependency through a sibling
working tree:

```toml
[patch.crates-io]
ct-dap-client = { path = "../codetracer/libs/ct-dap-client" }
```

A path dependency records no source, no checksum and no revision, so Cargo
copies the sibling's dependency edges verbatim into `Cargo.lock`. In this job
the sibling is **this repository's checkout**, so that lock hunk is a function
of the working directory, not of any pinned revision. `--locked` therefore
asserted "native-backend was relocked after the CodeTracer commit you are
testing" — something no CodeTracer author can satisfy from this repo — and
reported it by naming a flag.

The concrete instance: `libs/ct-dap-client` has declared `libc = "0.2"` since
`26f0bdba2` (the `cfg(unix)` `killpg` teardown in `DapStdioClient`);
native-backend's committed lock does not record that edge.

## What was verified before choosing a fix

With `codetracer@dev` (`52162408`) and `codetracer-trace-format@dev` checked out
as siblings of `codetracer-native-backend@dev` (`ed9088a1`), in an isolated
`CARGO_HOME`:

| command | result |
|---|---|
| `cargo fetch --locked` | exit 101, "needs to be updated but --locked was passed" |
| `cargo fetch` then `diff` | exit 0, **one line**: `+ "libc",` under `ct-dap-client` |

Nothing else in the lock moves. No package version, source or checksum changes.
The re-resolved lock is byte-identical to the head of
`codetracer-native-backend#83`.

## Why not the alternatives

**Make `ct-dap-client` a git dependency with a `rev`.** This would make the lock
self-contained, and it is the option the lock's shape superficially argues for.
Rejected: native-backend's flow tests drive *this repository's* `db-backend`
binary through *this repository's* `test_support` client. Pinning the client to
a revision while the binary is built from the sibling working tree makes the
harness disagree with the thing it drives. It also puts two actively developed
repositories on a mutual bump treadmill (native-backend pins codetracer;
codetracer's workspace pins native-backend), and it would have this repository's
own CI download a pinned copy of itself in order to test its HEAD. It would also
break the preflight's `locked-git-source-boundary` assertion, which exists to
keep exactly one git source — the private `lldb-sys` fork — in that lock.

**Keep the path dep for local development and resolve through `[patch]` for CI.**
Cargo has no conditional `[patch]`: whatever the path points at is what gets
locked. This is the status quo restated, not a fix.

**Stop checking in native-backend's `Cargo.lock`.** Rejected on evidence:
`codetracer-native-backend/nix/packages/default.nix` consumes it via
`cargoLock.lockFile`, so `build-nix`, `build-appimage` and `create-release` all
depend on it existing and being accurate.

**Add or remove `--locked` and stop there.** Removing it lets CI paper over
genuine version drift; adding it elsewhere blocks a developer legitimately
adding a dependency to `libs/ct-dap-client` until native-backend's lock moves.
A previous attempt declined this for the right reason.

## The change

Re-resolve the lock, then classify the difference. The comparison is *stricter*
than `--locked` in the dimension that matters and silent only where the answer
is genuinely not this repository's to give:

| difference | verdict |
|---|---|
| a package carrying a `source` changed version, source or checksum | **fail** — version drift, exactly what `--locked` was buying |
| a package carrying a `source` appears only after re-resolution | **fail** — a sibling pulled in a crate the lock does not pin, so Cargo took today's registry answer |
| a workspace member or in-repo path package changed its edges | **fail** — no sibling can explain it; the lock is unrefreshed |
| a *sibling* path package changed its edges | **report and continue**, naming package, sibling directory and revision |

The sibling checkout is restored byte-for-byte afterwards, so the gate still
never leaves another repository's lock rewritten.

### The failure is now legible

Today's `dev`, with the real siblings and the real Cargo:

```
codetracer-native-backend's Cargo.lock lags the sibling revisions under test.
  Every pinned version, source and checksum is unchanged; only the
  dependency edges of path packages differ, which is what a path
  dependency on a sibling working tree means:
    ct-dap-client (../codetracer/libs/ct-dap-client @ 521624086a00): +libc
  Refresh codetracer-native-backend's Cargo.lock against these siblings to clear
  this notice. The gate continues: nothing pinned moved.
```

And when something genuinely disagrees — committed lock pinned `zstd 0.12.4`
against a manifest requiring `0.13.3`, real `cargo fetch`, real repos:

```
codetracer-native-backend's Cargo.lock does not pin the graph it resolves to.
  A package that carries a source changed version, source or checksum:
    - zstd 0.12.4  (registry+https://github.com/rust-lang/crates.io-index)
    + zstd 0.13.3  (registry+https://github.com/rust-lang/crates.io-index)
  This is dependency version drift, not sibling lag. ...
```

## Mutation evidence

`ci/test/visual-replay-private-cargo-preflight-test.sh` gains four cases, one
per verdict. Each writes a real pair of lock files and runs the real preflight
against them; only the network fetch is stubbed. Every rule was verified to be
load-bearing by removing it and observing the suite go red:

| mutation | result |
|---|---|
| restore `--locked` on the fetch | suite fails (`run_positive_case` argv assertion) |
| guard invocation neutered | `sibling-lag` fails |
| version-drift rule removed | `version-drift` fails |
| new-pin rule removed | `new-pin` fails |
| local-member rule removed | `local-member-drift` fails |
| nothing removed | `Visual replay private Cargo preflight tests passed.` |

End to end against the real repositories: committed `dev` lock → passes with the
lag notice above; lock refreshed (`codetracer-native-backend#83`) → "matches the
sibling revisions under test"; `zstd` pin mutated → fails as shown.

`shellcheck` is clean on both scripts. `shfmt` diffs both files under the repo's
`.editorconfig`, and diffs the unmodified upstream copies identically — not a
regression from this change.

## Relationship to `codetracer-native-backend#83`

That PR restores the missing `libc` edge, and I verified independently that its
lock is byte-identical to a clean re-resolution against `codetracer@dev`. It is
correct and worth merging: it clears the notice this gate now prints. It is no
longer a *blocker*, and this PR deliberately does not depend on it landing —
after this change the gate is green on `dev` whether or not native-backend's
lock has been refreshed, because a stale sibling edge is not a CodeTracer
regression.

## Adjacent, not fixed here

`ci/test/beam-flow-cross-repo.sh` runs `cargo build --locked` against
`codetracer-beam-recorder`, which path-depends on `codetracer-trace-format` the
same way. The two `.github/workflows/codetracer.yml` BEAM build steps already
work around it by dropping `--locked` outright (see the comments at lines
~2230 and ~2276). Same class of defect, same available remedy; left for a
separate change.

### DIFF
--- a/ci/test/cargo-lock-pin-guard.py
+++ b/ci/test/cargo-lock-pin-guard.py
@@ -1,0 +1,259 @@
+#!/usr/bin/env python3
+"""Classify what changed when a sibling-coupled Cargo.lock is re-resolved.
+
+Why this exists
+---------------
+``codetracer-native-backend/Cargo.lock`` is *not* a pure function of that
+repository's own content.  Its manifest routes one dependency through a
+sibling working tree::
+
+    [patch.crates-io]
+    ct-dap-client = { path = "../codetracer/libs/ct-dap-client" }
+
+A path dependency records no source, no checksum and no revision, so Cargo
+writes the sibling's *dependency edges* into the lock verbatim.  The hunk is
+therefore a function of whatever happens to sit at ``../codetracer`` when the
+lock is written -- and in this gate that is the CodeTracer pull request under
+test, which is by design newer than any revision native-backend was locked
+against.
+
+That coupling is deliberate and correct: native-backend's flow tests drive the
+sibling's ``db-backend`` binary through the sibling's own ``test_support``
+client, so the client must come from the tree being tested, not from a pinned
+revision.  What is *not* correct is asserting the whole lock with ``--locked``:
+that turns an expected cross-repo lag into "cannot update the lock file because
+--locked was passed", a message that names a flag instead of a cause.
+
+So the gate re-resolves the lock and asks this script a sharper question than
+``--locked`` can: did anything that is actually pinned move?
+
+Verdicts
+--------
+FAIL  a package that carries a ``source`` changed version, source or checksum.
+      That is version drift, and it is exactly what ``--locked`` protected.
+FAIL  a package that carries a ``source`` is present only after re-resolution.
+      A sibling pulled in a crate the committed lock does not pin, so Cargo
+      just picked whatever the registry offers today.
+FAIL  a source-less package that is NOT reached through a sibling directory --
+      a workspace member, or an in-repo path dependency -- changed its edges.
+      Nothing outside the repository can explain that, so it is an unrefreshed
+      lock and the repository that owns it has to say so.
+PASS  differences confined to the dependency edges of packages reached through
+      a sibling directory (a path that leaves the repository), and to pinned
+      packages the re-resolved graph stopped using. This is cross-repo lag:
+      reported in full, with the sibling revision that caused it, so the next
+      reader sees the cause rather than the flag.
+"""
+
+from __future__ import annotations
+
+import argparse
+import sys
+import tomllib
+from pathlib import Path
+
+
+def load_packages(path: Path) -> list[dict]:
+    with path.open("rb") as handle:
+        return tomllib.load(handle).get("package", [])
+
+
+def pinned_identities(packages: list[dict]) -> dict[str, set[tuple]]:
+    """name -> {(version, source, checksum)} for packages that carry a source."""
+    out: dict[str, set[tuple]] = {}
+    for pkg in packages:
+        if "source" not in pkg:
+            continue
+        out.setdefault(pkg["name"], set()).add(
+            (pkg.get("version", ""), pkg["source"], pkg.get("checksum", ""))
+        )
+    return out
+
+
+def path_edges(packages: list[dict]) -> dict[str, list[str]]:
+    """name -> sorted dependency edges, for packages with no source (path deps)."""
+    return {
+        pkg["name"]: sorted(pkg.get("dependencies", []))
+        for pkg in packages
+        if "source" not in pkg
+    }
+
+
+def sibling_paths(manifest: Path) -> dict[str, str]:
+    """Map crate name -> declared relative path, for every path-routed entry."""
+    if not manifest.is_file():
+        return {}
+    with manifest.open("rb") as handle:
+        data = tomllib.load(handle)
+    found: dict[str, str] = {}
+    tables: list[dict] = []
+    for key in ("dependencies", "dev-dependencies", "build-dependencies"):
+        tables.append(data.get(key, {}))
+    tables.extend(data.get("patch", {}).values())
+    for target in data.get("target", {}).values():
+        for key in ("dependencies", "dev-dependencies", "build-dependencies"):
+            tables.append(target.get(key, {}))
+    for table in tables:
+        for name, spec in table.items():
+            if isinstance(spec, dict) and "path" in spec:
+                found.setdefault(name, spec["path"])
+    return found
+
+
+def describe_origin(repo: Path, relative: str | None) -> str:
+    """Human-readable '<path> @ <revision>' for a sibling crate directory."""
+    if relative is None:
+        return "unknown origin"
+    crate_dir = (repo / relative).resolve()
+    revision = "revision unavailable"
+    git_dir = crate_dir
+    for candidate in [git_dir, *git_dir.parents]:
+        if (candidate / ".git").exists():
+            head = candidate / ".git" / "HEAD"
+            try:
+                raw = head.read_text(encoding="utf-8").strip()
+            except OSError:
+                break
+            if raw.startswith("ref: "):
+                ref = (candidate / ".git" / raw[5:]).resolve()
+                try:
+                    revision = ref.read_text(encoding="utf-8").strip()[:12]
+                except OSError:
+                    revision = raw[5:]
+            else:
+                revision = raw[:12]
+            break
+    return f"{relative} @ {revision}"
+
+
+def main() -> int:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("--committed", required=True, type=Path)
+    parser.add_argument("--resolved", required=True, type=Path)
+    parser.add_argument("--manifest", required=True, type=Path)
+    parser.add_argument(
+        "--label",
+        default="codetracer-native-backend",
+        help="repository name used in the report",
+    )
+    args = parser.parse_args()
+
+    repo = args.manifest.parent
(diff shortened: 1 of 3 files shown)


zah:
## CI evidence: the lock defect is fixed; the gate now fails further downstream

`visual-replay-regression-gate` on this branch: run 31979123262, job 95243018513, runner `gpu-server-002-mcl-001`, raw log.

**The preflight passes.** The step that has reddened this gate since at least 2026-08-13 is gone. Both halves of the new contract are visible in the log:

```
23:28:19  codetracer-native-backend's Cargo.lock matches the sibling revisions under test.   (x3, the suite's own cases)
23:28:22  Visual replay private Cargo preflight tests passed.
23:28:29  codetracer-native-backend's Cargo.lock lags the sibling revisions under test.
            Every pinned version, source and checksum is unchanged; only the
            dependency edges of path packages differ, ...
              ct-dap-client (../codetracer/libs/ct-dap-client @ 005b4049ab3e): +libc
            Refresh codetracer-native-backend's Cargo.lock against these siblings to clear
            this notice. The gate continues: nothing pinned moved.
23:28:29  Running CodeTracer visual replay build prerequisites
```

`005b4049ab3e` is this PR's merge commit — the guard names the revision it actually compared against, which is the whole point of the change.

For comparison, `dev` tip (run 31975936283, job 95235318972) at the same point:

```
23:23:37  error: cannot update the lock file .../codetracer-native-backend/Cargo.lock because --locked was passed
23:23:37  Visual replay CI could not prefetch the native-backend Cargo graph.
```

The gate step ran 4 s longer than the whole failing step used to and reached `build-once`.

## The next blocker, which this PR does not fix

`build-once` → `tup` → the `ct` binary:

```
src/ct_test/process_exec.nim(25, 8) Error: cannot open file: runquota_process
 *** Command ID=6282 failed with return value 1
 *** tup: 1 job failed.
```

Mechanism, from the tree rather than the log:

* `src/ct/codetracer.nim:8` imports `../ct_test/ct_test`, so the `ct` binary now pulls in `src/ct_test/process_exec.nim`, whose first import is `runquota_process` (added by `a21d38a7`, extended by `bd11d761`).
* The runquota search paths live in **`src/ct_test/nim.cfg`** (relative sibling) and **`src/ct_test/config.nims`** (`RUNQUOTA_SRC` fallback). Nim reads `config.nims` from the *project* directory only, and `nim.cfg` from the project directory and its *parents*. Tup compiles `codetracer.nim` with the project directory at `src/ct`, so neither file is consulted — `src/ct_test` is a sibling of the project dir, not an ancestor.
* The repo-root `config.nims` does have a correct `RUNQUOTA_SRC` / `../runquota` block (lines 82–96), and the dev shell does export `RUNQUOTA_SRC` (`nix/shells/ci-base.nix:353`) — but the root `config.nims` is not read either, for the same reason. The tup build gets its entire search path from `NIM_REPO_PATH_FLAGS` in `src/Tuprules.tup` (lines 27–56), which has entries for `isonim`, `isonim-tui`, `nim-termctl`, `nim-pty`, `nim-everywhere`, `nim-acp`, `nim-agent-harbor`, `nim-agents` — and **none for runquota**.

Two independent gaps follow, both matching the shape of fixes already in the tree:

1. `src/Tuprules.tup` needs `--path:$(ROOT)/../runquota/libs/{runquota_process,runquota_core,runquota_host,runquota_host_linux,runquota_host_macos,runquota_host_windows,runquota_codec,runquota_protocol}/src`, exactly as `isonim-tui` and `nim-pty` are handled.
2. The sibling has to exist. This job's `setup-dev-env` `siblings:` list does not name `runquota`, and `scripts/require-siblings.sh` does not know about it either — which is why its preamble reported four missing optional siblings and not five. Adding it there is what makes the next occurrence say "runquota is missing" instead of "cannot open file: runquota_process".

This is present on `dev` too; it has simply been masked, because the gate died at the Cargo preflight before reaching `build-once`. This branch differs from `dev` only in three files under `ci/test/`, none of which participate in the Nim or tup build.

I have not attempted that fix here: it needs a tup + Nim + Nix dev-shell run to validate, and the only place that runs is this GPU-pinned job, so guessing at it would burn a scarce runner slot on an unverified change. Filing it as the next blocker rather than folding it in.

## Other checks

`workspace-lock-freshness` fails here and on `dev` tip with the identical cause — `No workspace lock exists in the shared manifests repo for any of:` — the manifests-repo publication outage. It is documented in the workflow as an alarm rather than a gate and is not in `ci/verdict/required-jobs.txt`.

Every other check on this run is still `queued`. That is not specific to this PR: `dev` runs 31975936283 (2026-08-16 22:17) and 31958696654 (2026-08-16 16:28) each have 22 jobs queued and only these same two ever reaching a conclusion. It is also not limited to the arm64 classes — `eph-linux-x64` jobs (`lint-bash`, `lint-nix`, `lint-rust`, `test-non-gui`) are queued alongside them. A possibly related observation, offered as a lead and not a conclusion: `GET /orgs/metacraft-labs/actions/runners` returns all eight `garm-*` ephemeral runners with `labels: []` and `runner_group: null`, and a runner with no labels can never match `runs-on: eph-linux-x64`.

zah:
Coordination note: #642 fixes the failure that surfaces immediately after this
one — `build-once` dying on
`src/ct_test/process_exec.nim(25, 8) Error: cannot open file: runquota_process`
(issue #641), which your branch was the first to expose.

The two touch disjoint files (#642: `src/Tuprules.tup`,
`scripts/require-siblings.sh`, the `siblings:` lists, and the two
`src/ct_test` config headers), so they can merge in either order — but the gate
cannot demonstrate #642 until this PR lands, because without it the job stops at
the `--locked` preflight. A run of #642 on top of plain `dev` confirmed exactly
that.

No change requested here.

PR #642, opened 2026-08-17
fix(build): declare runquota where the build actually looks for it

### LINKED ISSUE #641: visual-replay-regression-gate: tup build of `ct` cannot find `runquota_process`
## Symptom

`visual-replay-regression-gate` fails in `build-once`:

```
src/ct_test/process_exec.nim(25, 8) Error: cannot open file: runquota_process
 *** Command ID=6282 failed with return value 1
 *** tup: 1 job failed.
```

Observed on run 31979123262, job 95243018513 (raw log, 2026-08-16 23:28:35).

This is present on `dev`. It has been masked since at least 2026-08-13 because the gate died earlier, in `ci/test/visual-replay-private-cargo-preflight.sh`, on `cargo fetch --locked` against `codetracer-native-backend`'s Cargo.lock. #640 fixes that earlier failure and this became the next one.

## Mechanism

* `src/ct/codetracer.nim:8` imports `../ct_test/ct_test`, so the `ct` binary pulls in `src/ct_test/process_exec.nim`, whose first import is `runquota_process` (introduced by `a21d38a7`, extended by `bd11d761`).
* The runquota search paths live in `src/ct_test/nim.cfg` (relative sibling) and `src/ct_test/config.nims` (`RUNQUOTA_SRC` env fallback). Nim reads `config.nims` from the **project directory** only, and `nim.cfg` from the project directory and its **parents**. Tup compiles `codetracer.nim` with the project directory at `src/ct`, and `src/ct_test` is a *sibling* of that directory, not an ancestor — so neither file is consulted.
* The repo-root `config.nims` carries a correct block (lines 82–96: `RUNQUOTA_SRC`, then `../runquota/libs/<lib>/src`), and `nix/shells/ci-base.nix:353` does export `RUNQUOTA_SRC`. Neither helps: the root `config.nims` is not read for the same reason, and Nim `.cfg` files cannot expand environment variables in `--path` anyway.
* The tup build takes its entire Nim search path from `NIM_REPO_PATH_FLAGS` in `src/Tuprules.tup` (lines 27–56). That list covers `isonim`, `isonim-tui`, `isonim-gpui`, `nim-termctl`, `nim-pty`, `nim-everywhere`, `nim-acp`, `nim-agent-harbor`, `nim-agents` — and nothing for runquota.

## Suggested fix

Two parts, both matching patterns already in the tree:

1. **`src/Tuprules.tup`** — add the runquota libraries to `NIM_REPO_PATH_FLAGS`, exactly as `isonim-tui` / `nim-pty` are handled:

   ```
   --path:$(ROOT)/../runquota/libs/runquota_process/src\
   --path:$(ROOT)/../runquota/libs/runquota_core/src\
   --path:$(ROOT)/../runquota/libs/runquota_host/src\
   --path:$(ROOT)/../runquota/libs/runquota_host_linux/src\
   --path:$(ROOT)/../runquota/libs/runquota_host_macos/src\
   --path:$(ROOT)/../runquota/libs/runquota_host_windows/src\
   --path:$(ROOT)/../runquota/libs/runquota_codec/src\
   --path:$(ROOT)/../runquota/libs/runquota_protocol/src
   ```

2. **Provision the sibling, and diagnose it by name.** `visual-replay-regression-gate`'s `setup-dev-env` `siblings:` list does not include `runquota`, and `scripts/require-siblings.sh` does not know about it — which is why the gate's own preamble reported four missing optional siblings and not five:

   ```
   scripts/require-siblings.sh: 4 optional sibling repo(s) are missing.
     * isonim-tui  -- src/Tuprules.tup:48 puts it on the Nim search path
     * isonim-gpui -- src/Tuprules.tup:49 puts it on the Nim search path
     * nim-termctl -- src/Tuprules.tup:50 puts it on the Nim search path
     * nim-pty     -- src/Tuprules.tup:51 puts it on the Nim search path
   ```

   Adding `runquota=dev` to `siblings:` and a `runquota|libs/runquota_process/src||src/Tuprules.tup:NN ...` entry to `require-siblings.sh` makes the next recurrence say "runquota is missing" up front instead of surfacing as an opaque `cannot open file:` several minutes into the Nim compile — which is what `require-siblings.sh`'s header comment says it exists to do.

   `runquota` is already a workspace project and codetracer's flake already tracks `github:metacraft-labs/runquota/dev` (`flake.nix:108–111`).

Every other `--path` in that list points at a sibling checkout rather than at a nix-store input, so routing runquota the same way keeps one mechanism rather than adding a second. If the sibling checkout is not wanted here, the alternative is to make the tup rule consume `RUNQUOTA_SRC` (`export RUNQUOTA_SRC` is already the pattern used at `src/Tuprules.tup:387–413` for `CARGO_HOME`, `NIMFLAGS`, and friends) — but that needs the paths expanded before they reach `nim`, since `--path` does not expand environment variables.

Not attempted in #640: validating it needs a tup + Nim + Nix dev-shell run, and the only place that runs is this GPU-pinned job, so an unverified guess would burn a scarce runner slot.

zah:
Fixes #641. Sits on top of #640 in the `visual-replay-regression-gate` failure
queue: #640 clears the private-Cargo preflight, this clears the `build-once`
step immediately after it. The two touch disjoint files and can land in either
order.

## What actually causes it (the issue's mechanism is partly wrong)

#641 says the runquota paths are unreachable because "Nim reads `config.nims`
from the project directory only", so the repo-root `config.nims` is not
consulted. That is not right, and it matters: Nim reads `config.nims` from the
project directory **and its ancestors**, so the repo-root file *is* read for a
`src/ct` project. Verified directly:

```
$ nim dump --dump.format:json src/ct/codetracer.nim   # runquota checked out
/home/zahary/m/repro-fixes/runquota/libs/runquota_process/src
... 8 runquota paths present
```

The half of the mechanism that holds is the **sibling** half: `src/ct_test/nim.cfg`
and `src/ct_test/config.nims` are not ancestors of `src/ct`, so they contribute
nothing to `ct`.

So the real cause is narrower, and it is not "no declaration exists" but "every
declaration that exists needs something CI does not provide":

| source | status for the tup `ct` compile |
|---|---|
| `src/ct_test/nim.cfg` / `config.nims` | not read — sibling directory, not an ancestor |
| repo-root `config.nims`, `RUNQUOTA_SRC` branch | read, but tup sanitizes the environment to the `export` list in `src/Tuprules.tup`, which does **not** name `RUNQUOTA_SRC` |
| repo-root `config.nims`, `../runquota` sibling branch | read, and works — **but nothing checks the sibling out** |
| `src/Tuprules.tup` NIM_REPO_PATH_FLAGS | no runquota entry at all |

The load-bearing omission is therefore the missing **checkout**: the gate's
`siblings:` list never asked for `runquota`. Adding the `--path` flags alone
would not have fixed CI, because they point at `$(ROOT)/../runquota` too.

## Changes

1. `src/Tuprules.tup` — the eight runquota library paths on NIM_REPO_PATH_FLAGS,
   so the tup build states its own dependency rather than inheriting it from a
   config whose environment half is stripped.
2. `scripts/require-siblings.sh` — `runquota` in the **required** tier. Override
   column deliberately empty: accepting `RUNQUOTA_SRC` would green the check on
   the strength of a variable tup discards, and hand the job back the same
   `cannot open file`.
3. All four jobs that reach `just build-once` provision it —
   `visual-replay-regression-gate` and `test-ui-tests` via `siblings:`,
   `viewmodel-tests` and `cross-process-linux` via an explicit clone step
   (neither of their composite actions owns this dependency).
4. `src/ct_test/nim.cfg` and `config.nims` keep a scope note. They are **not**
   dead — `ci/test/m16-release-gate.sh` compiles `src/ct_test/*.nim`, making
   `src/ct_test` the project directory, and there they are load-bearing. They
   are only silent for `ct`. The note records that boundary, since a file that
   looks authoritative and is conditionally ignored is how this stayed hidden.

## Validation (local, in the dev shell)

Reproduced and fixed on the real import chain, project dir `src/ct`, configs
disabled so only NIM_REPO_PATH_FLAGS is in play:

```
BEFORE (flags as on dev): src/ct_test/process_exec.nim(25, 8) Error: cannot open file: runquota_process
AFTER  (flags as in this PR): binary built, runquota_process resolved
```

The `BEFORE` line is byte-identical to what the gate reported.

`just build-once` was run end to end. The `ct` compile it now emits carries the
runquota paths and produces **zero** `cannot open file` errors. It does not go
fully green on this host for two pre-existing, unrelated reasons: 8 Nim targets
fail on `could not load: libpcre.so` (the dev shell exposes pcre only through
`LD_LIBRARY_PATH`, which tup strips — it is not on the `export` list), and
`codetracer_trace_writer_nim`'s build script exits 101. Neither is touched by
this change and neither occurs on the CI runner, which got past every Nim
compile to reach the runquota error.

Mutation check on the new prerequisite:

```
runquota present -> exit 0
runquota removed -> exit 1
  * runquota
      expected:  <workspace>/runquota/libs/runquota_process/src/runquota_process.nim
      state:     not checked out
```

`ci/lint/bash.sh` and `scripts/test-build-alignment.sh` (63 checks) pass.

### DIFF
--- a/.github/workflows/codetracer.yml
+++ b/.github/workflows/codetracer.yml
@@ -1477,6 +1477,23 @@ jobs:
         with:
           gh-token: ${{ steps.ci_token.outputs.token }}
 
+      - name: Clone runquota sibling
+        # `just build-once` below compiles `ct`, and src/ct/codetracer.nim ->
+        # ../ct_test/ct_test -> src/ct_test/process_exec.nim opens with
+        # `import runquota_process`. On the tup driver this checkout is the only
+        # source of those modules: src/Tuprules.tup's NIM_REPO_PATH_FLAGS names
+        # `$(ROOT)/../runquota/...`, and tup's environment allowlist in that
+        # same file omits RUNQUOTA_SRC, so the dev shell export cannot stand in
+        # for it. Cloned here rather than via the two composite actions above
+        # because neither owns this dependency (issue #641).
+        uses: metacraft-labs/metacraft-github-actions/clone-repo@main
+        with:
+          repo: metacraft-labs/runquota
+          ref: dev
+          path: ${{ github.workspace }}/../runquota
+          gh-token: ${{ steps.ci_token.outputs.token }}
+          submodules: 'false'
+
       - name: Build frontend (needed by gated Playwright stage)
         env:
           CODETRACER_DB_BACKEND_SKIP_DIRENV: "1"
@@ -1597,6 +1614,21 @@ jobs:
         # zero-test guard that fails if they all skip (so a missing recorder
         # cannot masquerade as a pass).
         run: nix develop .#devShells.x86_64-linux.default --command just test-vm-recorder-gated
+
+      - name: Clone runquota sibling
+        # The `just build-once` step below compiles `ct`, whose import chain
+        # (src/ct/codetracer.nim -> ../ct_test/ct_test ->
+        # src/ct_test/process_exec.nim) starts with `import runquota_process`.
+        # Under tup the checkout is the only way those modules resolve --
+        # NIM_REPO_PATH_FLAGS points at `$(ROOT)/../runquota/...` and tup's
+        # environment allowlist omits RUNQUOTA_SRC (issue #641).
+        uses: metacraft-labs/metacraft-github-actions/clone-repo@main
+        with:
+          repo: metacraft-labs/runquota
+          ref: dev
+          path: ${{ github.workspace }}/../runquota
+          gh-token: ${{ steps.ci_token.outputs.token }}
+          submodules: 'false'
 
       - name: Build ct (needed by the CLI dispatch tests)
         # record_missing_recorder_test.nim and record_dispatch_e2e_test.nim
@@ -2352,6 +2384,14 @@ jobs:
         # (setup-nix). Nix is provisioned here for both the macOS sibling
         # ``nix develop`` builds and the ct build + Playwright run, which now
         # all go through codetracer's dev shell.
+        #
+        # ``runquota`` is listed because the macOS leg runs ``just build-once``
+        # and `ct` imports ``runquota_process`` through
+        # src/ct_test/process_exec.nim. On Darwin build-once selects the
+        # reprobuild driver, which could resolve runquota from its nix flake
+        # input alone -- but scripts/require-siblings.sh requires the checkout
+        # on every driver so there is a single workspace-completeness rule (the
+        # tup driver has no other way in; see the notes in that script).
         uses: metacraft-labs/metacraft-github-actions/setup-dev-env@main
         with:
           env-flavor: nix
@@ -2366,6 +2406,7 @@ jobs:
             nim-stackable-hooks=dev
             codetracer-visual-replay=dev
             codetracer-native-backend=dev
+            runquota=dev
 
       - name: Stage codetracer-trace-format into libs/
         # The nix flake build below overrides the codetracer-trace-format input
@@ -2744,6 +2785,16 @@ jobs:
         # frontend's IsoNim/agent libraries, the incremental runner's trace and
         # I/O libraries, and visual replay's own native siblings. A missing
         # lock entry fails loudly before the build starts.
+        #
+        # `runquota` is on the list because this gate runs `just build-once`,
+        # which compiles `ct`, which pulls in src/ct_test/process_exec.nim ->
+        # `import runquota_process`. Under the tup driver this CHECKOUT is the
+        # only way those modules resolve: NIM_REPO_PATH_FLAGS points at
+        # `$(ROOT)/../runquota/...`, and tup's environment allowlist in
+        # src/Tuprules.tup does not carry RUNQUOTA_SRC, so the dev shell's
+        # export of it never reaches `nim`. Its absence is what made this job
+        # die on `cannot open file: runquota_process` (issue #641), fifteen
+        # minutes into the Nim compile rather than here.
         uses: metacraft-labs/metacraft-github-actions/setup-dev-env@main
         with:
           env-flavor: nix
@@ -2767,6 +2818,7 @@ jobs:
             codetracer-native-recorder=dev
             codetracer-native-test-programs=dev
             codetracer-visual-replay-fixtures=dev
+            runquota=dev
 
       - name: Run visual replay regression gate
         env:
--- a/scripts/require-siblings.sh
+++ b/scripts/require-siblings.sh
@@ -11,7 +11,7 @@
 # sits next to `<workspace>/isonim`, `<workspace>/io-mon`, ... and both build
 # drivers reach across by relative path:
 #
-#   * Nim   -- `src/Tuprules.tup:47-55` passes `--path:$(ROOT)/../<sibling>/src`
+#   * Nim   -- `src/Tuprules.tup:72-88` passes `--path:$(ROOT)/../<sibling>/src`
 #              unconditionally, and `config.nims` adds the same directories
 #              through `addPathIfDir`.
 #   * Cargo -- `src/db-backend/Cargo.toml` declares `path = "../../../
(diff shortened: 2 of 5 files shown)


zah:
## CI result on this branch, and why it does not yet show the fix

The gate ran (job 95248001480) and failed — but **before** reaching the code this
PR changes, at the blocker #640 fixes:

```
```

This branch is cut from `dev`, which does not contain #640, so `build-once` is
never reached. That is **parity with `dev` tip**, not a regression: the run #641
cites as evidence (31979123262) was on `ci/visual-replay-lock-guard` — #640's own
branch — which is exactly why the runquota error was visible there and is not
visible here.

Ordering: **#640 must land first.** Once it does, this PR's effect becomes
observable at `build-once`. The two change disjoint files, so there is no
conflict either way.

What the run does confirm is that the provisioning half landed. The sibling is
now cloned and exported:

```
metacraft-labs/runquota -> dev (override)
CT_SIBLING_PATHS: ... runquota=/var/lib/github-runner-work/mcl-002/codetracer/runquota
```

Before this PR, `runquota` appeared nowhere in that list, and
`require-siblings.sh` reported four missing optional siblings rather than naming
this one.

The compile-level evidence is in the PR body: on the real import chain with
configs disabled, the pre-change flags reproduce
`src/ct_test/process_exec.nim(25, 8) Error: cannot open file: runquota_process`
byte-for-byte, and the post-change flags build.

zah:
Rebase onto `dev` — #640 is merged (`00:22Z`), so the `--locked` preflight defect that was stopping this branch's gate before it reached `build-once` is gone. The next `visual-replay-regression-gate` run on this PR should be the first that can actually demonstrate the runquota fix end to end.

zah:
Merging without waiting for CI. Recording what this rests on — including a correction to the framing.

**CI gives no evidence here, and I want to be explicit about that.** The premise that `visual-replay-regression-gate` fails on `dev` at `cannot open file: runquota_process` was **not true of the baseline run**. `grep -i "cannot open file"` returns zero hits in both the baseline log (job 95235318972) and this PR's own log (job 95248001480). Both died ~50 log lines earlier, at the private-Cargo preflight, byte-for-byte identically:

```
```

That is two absences, not a comparison, so I did not rely on it. The `runquota_process` error is real but only appears on [run 31979123262](https://github.com/metacraft-labs/codetracer/actions/runs/31979123262) (branch `ci/visual-replay-lock-guard` = #640), the only run that got past the preflight.

**This matters for sequencing, and it now resolves in this PR's favour:** #640 merged on 2026-08-17, and `dev` tip no longer passes `--locked` to that fetch. So the gate will now advance past the preflight and hit the runquota error on the next run — making this the live blocker rather than a speculative one.

**What I relied on instead: direct local reproduction, run twice independently.** Extracted the `--path` flags from `NIM_REPO_PATH_FLAGS` in `src/Tuprules.tup` with `$(ROOT)` expanded to the repo root (`ROOT = $(TUP_CWD)/../../`), unset `RUNQUOTA_SRC`, and skipped every config file (`--skipUserCfg --skipParentCfg --skipProjCfg`) to emulate what tup actually hands the compiler:

```
total=37 runquota=8   (all 8 --path targets verified to exist in the sibling)

=== BEFORE (dev: runquota paths removed) ===
imp.nim(1, 8) Error: cannot open file: runquota_process

=== AFTER (this PR's paths) ===
62812 lines; 4.253s; ... [SuccessX]
```

The BEFORE line is character-identical to the CI failure at `run640_vrrg.log:2608`. Because the test ran with `RUNQUOTA_SRC` unset and still succeeded, it also proves the fix does not depend on that variable.

**Correction to the mechanism as described.** The claim that `RUNQUOTA_SRC` "is now declared where the build looks" is not what the diff does. `src/Tuprules.tup` has exactly 16 `export` lines before *and after* this change, and `RUNQUOTA_SRC` is in neither list. The PR routes around the export list with literal `--path` flags instead — which is the right call, since `--path` cannot expand environment variables anyway. The *stripping* half of the original claim is true; the *"now added"* half is not, and the code comment is honest about this ("accepting `RUNQUOTA_SRC` would green the check on the strength of a variable tup discards").

**Also worth stating:** the sibling checkout is load-bearing only *given* the new `--path` flags. The repo-root `config.nims` would have resolved runquota on its own had the checkout existed — as it does locally, where `nim dump` shows 16 runquota entries in `lib_paths`. The `--path` flags are the belt to the checkout's braces, not the reverse. The earlier hypothesis that Nim would not read a repo-root `config.nims` is indeed wrong: `codetracer-nim/compiler/nimconf.nim:278-290` walks `parentDirs(..., fromRoot=true)` running `DefaultConfigNims` in every ancestor, on identical footing with `nim.cfg`.

**Coverage check:** all four jobs that reach `build-once` (`test-ui-tests`, `visual-replay-regression-gate`, `cross-process-linux`, `viewmodel-tests`) gained a `runquota` checkout; nothing in `nix/` or `flake.nix` invokes `build-once.sh`, so the new hard requirement in `scripts/require-siblings.sh` has no uncovered caller. Rebases cleanly onto current `dev`.

**Badges flipping green -> red:** none — `visual-replay-regression-gate` is already red.

zah:
Post-merge confirmation from production, plus one caveat worth a follow-up.

**The fix works.** `visual-replay-regression-gate` on `dev` @ `85c6dfe73` (job [95867630283](https://github.com/metacraft-labs/codetracer/actions/runs/32185366280)):

- The sibling is now provisioned: `metacraft-labs/runquota -> dev (override)`, and `runquota=/var/lib/github-runner-work/mcl-002/codetracer/runquota` appears in `CT_SIBLING_PATHS`.
- The `nim` command lines now carry all eight `--path:../../..//../runquota/libs/runquota_*/src` flags.
- **`grep -c "cannot open file: runquota_process"` -> 0.**
- **tup reaches `100%`** and completes. It previously died at `*** tup: 1 job failed`.

The controlled comparison is as clean as it gets: the immediately preceding `dev` commit `ac39e9d05` — which carries #632/#635/#636/#637/#638/#639 but *not* this PR — fails on the same runner class with exactly `src/ct_test/process_exec.nim(25, 8) Error: cannot open file: runquota_process` and `tup: 1 job failed`. One commit apart, error gone, build completes.

**The job is still red, ~15 minutes further along, for an unrelated reason:**

```
building /nix/store/chcglhq99875iln32hr1wphk1d1ni44l-nim-2.2.8.drv...
building /nix/store/796yldz1mx5wd7z5brd6sxxbfkha315n-nix-shell-env.drv...
bash: line 2: nimble: command not found
error: Recipe `test-visual-replay-gate` failed on line 674 with exit code 127
```

That is a *different* dev shell (it is building `nim-2.2.8` from source rather than reusing codetracer's) whose environment lacks `nimble`. New failure, new cause, worth its own issue — not a regression from this change.

---

**Caveat this exposed, and I think it should be followed up.** The gate now emits:

> `##[warning]sibling 'runquota' is pinned by the explicit override 'runquota=dev', not by the workspace lock, and 'dev' is not a 40-hex commit SHA. This build is therefore not reproducible: it resolves to whatever that ref points at right now.`

That warning is correct. The `siblings:` entries added here use `runquota=dev`, which follows the existing convention in these workflows, but it means the gate now builds against a moving target. The durable fix is to declare `runquota` in codetracer's project manifest so the workspace lock can pin it, after which the `=dev` overrides can be dropped. I have not done that — the manifest repo was outside this pass's scope — but it is the natural next step and this warning will keep firing until it happens.

PR #655, opened 2026-08-26
fix(build): report why runquotad failed to start instead of just a log path

zah:
When `runquotad` fails to come up, `build-once.sh` printed only a path to a
log file and exited. In CI that log is never uploaded, so the one artefact
that identifies the cause was discarded at the moment of failure — every
affected run reported the same opaque message and nothing else.

Both failure paths now report, to stderr: the runquotad binary that was
selected and how it was resolved, its `--version`, the socket path together
with its length in bytes, the exact argument list used, and the full
contents of the daemon log, clearly delimited. Missing or empty logs are
called out explicitly rather than printing nothing. The "binary not found"
path lists every candidate location it searched.

The socket length is reported because an over-long path is a real failure
mode: an AF_UNIX socket path is limited to 108 bytes, and the socket is
derived from `TMPDIR`, so a long temporary directory makes the daemon exit
immediately with a message that otherwise looks like an unexplained
startup failure.

The success path is unchanged and emits no additional output; this was
verified by comparing stdout and stderr byte for byte against the previous
version with a real daemon. The version probe runs only on failure paths
and is bounded so a hung daemon cannot stall the report.

### DIFF
--- a/scripts/build-once.sh
+++ b/scripts/build-once.sh
@@ -220,44 +220,141 @@ if [ -n "$ct_reprobuild_host" ]; then
 		fi
 	fi
 
+	# Ask the daemon binary to identify itself. `runquotad --version` prints its
+	# version and exits, but only when it is the SOLE argument -- with anything
+	# else alongside it the process goes on to serve -- so pass nothing else.
+	# The call is run detached behind a bounded wait rather than through
+	# `timeout(1)`, which is not available on every host this script supports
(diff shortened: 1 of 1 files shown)


zah:
Superseded by ca1be6341, which is already on dev and fixes this properly: it moves the socket into a per-run directory the daemon creates with the mode it requires, and prints the daemon log on both failure branches. This PR was written against a pre-ca1be6341 checkout and duplicates it.

PR #659, opened 2026-08-26
fix: unblock the dev shell (runquota pin) and stop the direnv misdiagnosis that hid it

zah:
`dev` has produced no compiler run for days. Two defects, one behind the other.

## 1. The dev shell does not build: `runquota` is not the revision reprobuild is compiled against

`flake.nix` sets `inputs.runquota-src.follows = "runquota"`, so the pinned
reprobuild is COMPILED against whatever this repo's `runquota` input resolves
to. `reprobuild` is pinned to an exact SHA; `runquota` tracked the `dev`
branch. The two drifted.

c9cb186c mirrored the nixos-modules reprobuild pin to `b5b88139` without moving
`runquota`, which sat at `f3382f5` (2026-07-04). reprobuild@b5b88139's own
flake.lock pins `runquota-src` at `b71e8e9`. `nix develop` died with

```
libs/repro_runquota/src/repro_runquota.nim(932, 18)
  Error: undeclared identifier: 'ExtensionCellWire'
```

on a tree whose own sources were fine. Observed in CI (run 33021054045) and
reproduced locally at the identical derivation hash
`1r4xm2mvn6gydmm127a8j4j261v5fzs0-reprobuild-0.1.3.drv`.

Freshening `runquota` to its `dev` tip is **not** the fix: `dev` had since
REMOVED a type, and the tip merely trades one error for its mirror image
(`undeclared identifier: 'LeaseFinishOutcome'`, same file, line 501). Both
directions were observed here. The pin is now the exact SHA reprobuild locks.

`scripts/test-flake-pin-alignment.sh`, wired into `just test`, asserts the
equality from two flake.lock files — no toolchain, no network, well under a
second. It distinguishes BEHIND from AHEAD, and when the sibling `reprobuild`
checkout it must read is absent it SKIPS LOUDLY rather than reporting a pass it
did not earn.

## 2. The gate blamed `direnv` for a shell that never built

`nix develop <shell> --command <probe>` exits non-zero both when the shell fails
to BUILD and when the probe fails. The "Allow direnv in the freshly cloned
siblings" step funnelled both into `direnv is not on PATH in the codetracer ci
dev shell` — a claim that is false, and provably so from the step's own comment
four lines above, which correctly states direnv is declared in
`nix/shells/ci-base.nix`. In run 33021054045 that message sat directly beneath
nix's `Error: undeclared identifier: 'ExtensionCellWire'`.

The shell is now built on its own first and a failure there is reported as a
shell build failure. `ci/test/launcher-recorder-e2e-workflow-test.sh` extracts
the step's script from the YAML and RUNS it against a stub `nix`, once per
failure mode; reverting the step makes two of the three new assertions fail.

## Verification

Both changes are red-test-first and mutation-tested; the mutation tables are in
the commit messages and the test headers. `just test-flake-pin-alignment` and
`bash ci/test/launcher-recorder-e2e-workflow-test.sh` (18/18) pass.

### DIFF
--- a/.github/workflows/launcher-recorder-e2e.yml
+++ b/.github/workflows/launcher-recorder-e2e.yml
@@ -537,6 +537,26 @@ jobs:
         # PATH inside the shell used here; the check below turns a future
         # removal into a named failure instead of a bare "direnv: not found"
         # exit 127 deep inside the driver.
+        #
+        # THE TWO FAILURE MODES ARE SEPARATED ON PURPOSE, and the reason is a
+        # real misdiagnosis this step produced. `nix develop <shell> --command
+        # <probe>` exits non-zero for two unrelated reasons: the shell did not
+        # BUILD, or it built and the probe failed. A single `|| { echo direnv
+        # is not on PATH; }` reports the second for both.
+        #
+        # In run 33021054045 the `ci` shell failed to build — reprobuild would
+        # not compile against the runquota revision this repo pinned — and this
+        # step announced "direnv is not on PATH in the codetracer ci dev shell"
+        # while nix's own `Error: undeclared identifier: 'ExtensionCellWire'`
+        # scrolled past above it. The claim was not merely unhelpful, it was
+        # FALSE, and provably so from this file: the comment four lines up
+        # states direnv is declared in ci-base.nix, and it is. Anyone trusting
+        # the error would have gone looking for a missing package in a shell
+        # that never existed.
+        #
+        # So the shell is built first, on its own, and a failure there is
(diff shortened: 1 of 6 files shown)


zah:
Landed on `dev` via rebase-merge as 6a8836468 and 8b3154fbc — identical content, confirmed by patch-id (390afd86d4c9 and 9a714c2ebff6 on both sides). GitHub did not auto-close it because the SHAs differ after the rebase.

The dev-shell fix this carried unblocked 8 of the 12 failing jobs on `dev`.

PR #660, opened 2026-08-26
ci: route the Linux jobs to the GPU-host scale sets (rr jobs held back)

zah:
Moves 25 Linux job legs off `high-mem-server` onto `eph-linux-x64-g1` / `-g2`.

`codetracer` is **46.34% of clean job-minutes** on the `eph-linux-x64` class — 3.2× `vm-harness`, 4.9× `nixos-modules`. `high-mem-server` is a Dell R620 on six consumer QLC SSDs that sustain ~78 MB/s; four watchdog resets in 48h opened the campaign this belongs to. The GPU scale sets are live and proven: each ran ~105 container creates in the last 24h after earlier routing landed.

## ⚠ The rr jobs are deliberately NOT routed, and the reason is measured

`test-ui-tests-rr` and `rr-backend-tests` stay on `eph-linux-x64`.

**The container is not the obstacle** — the guest reports `perf_event_paranoid = -1`, exposes `cpu_core`/`cpu_atom`/`intel_pt`, and an unprivileged `perf_event_open(HW_BRANCH_INSTRUCTIONS)` probe succeeds at `CapEff: 0`.

**The CPU is.** Both GPU hosts are **i9-12900K (Alder Lake), hybrid**: `cpu_core` = CPUs 0-15, `cpu_atom` = 16-23. Only the P-cores implement `r5111c4`, the retired-conditional-branch counter rr requires. Measured on both hosts:

| case | -g1 | -g2 |
|---|---|---|
| pinned P-core | record+replay OK | record+replay OK |
| pinned E-core | abort in `PerfCounters::start()` | abort |
| **unpinned** (how CI runs rr) | **OK** | **FAILED** |

Unpinned is a per-job coin flip — the identical command passed on `-g1` and failed on `-g2` in a single run. Routing as-is manufactures flake. They can move the moment rr is confined to CPUs 0-15 (taskset/affinity on the runner, or an E-core-free class).

Two corrections found while measuring: **`cross-process-linux` has no rr references** (`scripts/test-cross-process.sh` is db-backend only), so it *is* routed; and **rr is not actually exercised in CI today** — both rr jobs are red before reaching a counter (`CODETRACER_RR_BACKEND_PATH is not set`; `libs/rr/flake.nix ... not tracked by Git`).

## Split — balanced by runtime, not job count

Averages over the last 51 completed `dev` runs, cancelled/skipped excluded. Almost every routed job declares no `needs:`, so ~17 Linux jobs queue at t=0 against 8 slots and the balance sets the critical path.

- **`-g1` (13)**: ct-test-providers 31.9, appimage-build 21.7, dev-build 20.9, viewmodel-tests 19.1, cross-process-linux 15.0, lint-bash 14.5, test-non-gui(nixos) 12.3, reprobuild-linux-smoke 7.0, appimage-lib-check, push-to-attic, shell-recorder-tests, sidecar-retirement, beam-flow-dap
- **`-g2` (12)**: test-ui-tests(nixos) 29.4, lint-rust 28.1, nix-build 17.4, lint-nix 16.5, build-and-deploy-docs 15.0, lint-nim 14.3, ct-test-release-gate 11.1, push-gpg, push-install, push-tag, create-release, launcher-recorder-e2e

Balanced across all three load regimes: full run **142.4 vs 132.5**, always-on subset **99.8 vs 100.1**, frequency-weighted **115.5 vs 114.5**. `launcher-recorder-e2e` is on `-g2` because desktop-edge calls it 3× concurrently, so `-g2` carries the lighter build wave.

Split across **both** hosts deliberately: 46% of runtime into one host's 4 slots would just relocate the bottleneck onto machines whose documented failure mode is a silent multi-hour freeze where the watchdog has been observed not to fire. **`maxRunners` is untouched at 4.**

## Expect the first wave to be slower

The GPU hosts' `/nix/store` has never built codetracer, so the first run pays a full Attic fetch. That is a one-off, not a regression.

## Verified

No branch protection or required checks on `dev`/`stable`; the "Protect main" ruleset is default-branch-only. `required-jobs.sh` reads job **ids** from `toJSON(needs)`, so the matrix rename does not break the verdict gate. Shared-store sweep re-confirmed: no hits for `/srv/repro-store` or `/var/lib/reprobuild/shared-store`. All 10 workflow files re-parse clean with `yq`.

Targets **`dev`**, not the `stable` default: `origin/stable...origin/dev` is 0 behind / 188 ahead.

### DIFF
--- a/.github/workflows/beam-flow.yml
+++ b/.github/workflows/beam-flow.yml
@@ -58,7 +58,7 @@ permissions:
 
 jobs:
   beam-flow-dap:
-    runs-on: eph-linux-x64  # CIP-5: nix job → eph-linux-x64
+    runs-on: eph-linux-x64-g1  # CIP-5 nix job; H4: routed to GPU host gpu-server-001
 
     steps:
       - name: Generate CI token
--- a/.github/workflows/codetracer.yml
+++ b/.github/workflows/codetracer.yml
@@ -1123,7 +1123,7 @@ jobs:
             -E 'test(~dap) and not test(~bash_flow) and not test(~zsh_flow) and not test(~javascript_flow)'
 
   lint-bash:
-    runs-on: eph-linux-x64  # CIP-5: nix job (self-hosted nixos) -> eph-linux-x64
+    runs-on: eph-linux-x64-g1  # CIP-5 nix job; H4: routed to GPU host gpu-server-001
     steps:
       - name: Mint installation token for cross-repo sibling access
         # GitHub free plan restricts org-level secrets with visibility
@@ -1161,7 +1161,7 @@ jobs:
       - run: "nix develop .#devShells.x86_64-linux.default -c ./ci/lint/bash.sh"
 
   lint-nim:
-    runs-on: eph-linux-x64  # CIP-5: nix job (self-hosted nixos) -> eph-linux-x64
+    runs-on: eph-linux-x64-g2  # CIP-5 nix job; H4: routed to GPU host gpu-server-002
     steps:
       - name: Mint installation token for cross-repo sibling access
         # GitHub free plan restricts org-level secrets with visibility
@@ -1199,7 +1199,7 @@ jobs:
       - run: "nix develop .#devShells.x86_64-linux.default -c ./ci/lint/nim.sh"
 
   lint-nix:
-    runs-on: eph-linux-x64  # CIP-5: nix job (self-hosted nixos) -> eph-linux-x64
+    runs-on: eph-linux-x64-g2  # CIP-5 nix job; H4: routed to GPU host gpu-server-002
     steps:
       - name: Mint installation token for cross-repo sibling access
         # GitHub free plan restricts org-level secrets with visibility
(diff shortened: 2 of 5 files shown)


zah:
## Correction — the rr jobs should NOT have been held back

The operator pushed back on this and was right. **rr handles hybrid CPUs itself**, and my analysis missed it by testing the guest instead of reading rr.

From the vendored source in `codetracer-rr`:

- `RecordCommand.cc:214` — the **default** bind mode is `BindCPU::PREFER_PERF_CORE`.
- `util.cc:2165` — `choose_cpu()` calls `filter_for_perf_cores(cpus)` before selecting.
- `util.cc:2075` — that filter keeps only CPUs in `CPUGroup::P_CORE`.
- `PerfCounters.cc:216` — rr's own comment: *"For hybrid products the data in this table is for the P-cores of that product."*

So on a 12900K rr already confines itself to CPUs 0–15. The E-core abort I measured is real but **only reachable by explicitly binding to an E-core**, which is precisely what my probe did — the failure was manufactured by the test.

The reported "unpinned failed on -g2" result does not survive either: `PerfCounters.cc:325` makes `UNBOUND` a hard fatal on multi-PMU systems (*"Multiple PMU types detected. Unbinding CPU is not supported"*), and nothing in codetracer's CI passes `--bind-to-cpu=any`, so `UNBOUND` is never reached. Whatever my unpinned run did, it was not the code path CI takes.

**`test-ui-tests-rr` and `rr-backend-tests` should be routed with everything else.** I'll push that change to this branch.

The rest of the PR stands unchanged — the split, the balance figures, and the base-branch verification are unaffected.

One finding from that work does survive independently: **rr is not currently exercised in CI at all.** Both rr jobs are red before reaching any counter (`CODETRACER_RR_BACKEND_PATH is not set`; `libs/rr/flake.nix ... not tracked by Git`). Worth fixing separately — routing them is necessary but not sufficient for them to actually test anything.

PR #661, opened 2026-08-27
fix(ci): make Cross-Repo Integration Tests able to pass on dev

zah:
> **Depends on #659.** Every job in this workflow enters
> `nix develop .#devShells.x86_64-linux.default`, and on `c9cb186c` that shell
> does not build — `reprobuild@b5b88139` will not compile against the runquota
> revision this repo pins (`Error: undeclared identifier: 'ExtensionCellWire'`).
> #659's branch is merged in here so this lane can be demonstrated at all; when
> #659 lands on `dev`, this PR's diff shrinks to its own four commits.

`Cross-Repo Integration Tests` has never been green on `dev` — not once in the
last 200 runs of the workflow, on any branch. Both of its jobs fail for reasons
that are visible in the workflow file, and both fail *late*, after twenty
minutes of compilation, because nothing checks the precondition they violate.

### `shell-recorder-tests`

It runs `cargo test` from inside `src/db-backend`, so it compiles
`src/db-backend/build.rs`, which requires the `codetracer-native-recorder`
sibling unconditionally — `src/lib.rs` exports `emulator_ffi`,
`emulator_origin`, `emulator_session` and `data_watch` whether or not anything
calls them. Its hand-rolled `siblings:` list was derived from the shell
recorders' own path dependencies and did not name it, so every run since
2026-08-04 died at `build.rs:170`.

The job now delegates to `.github/actions/setup-db-backend-siblings` — which is
where that set is defined once, for the eight jobs that build the crate —
through a new `extra-siblings` input, and gains the `direnv allow` /
`nimble install --depsOnly` / `CT_EMULATOR_EXTRA_NIM_PATHS` preamble every other
db-backend leg carries.

### `rr-backend-tests`

Two defects, in the order they fire.

1. It died four steps after an unfold step that reported success having unfolded
   one of two submodules. The skipped one was *uninitialised* — a state the
   sibling clone can legitimately leave behind, since it fetches submodules with
   `--submodules-optional` — and the loop had no diagnostic for that branch. It
   is intermittent: run 32489213400 unfolded both. The logic moved to
   `ci/unfold-submodules.sh`, which repairs that state, enumerates before
   mutating, and verifies that the paths `codetracer-native-backend`'s `path:`
   flake inputs need are actually tracked.
2. The step after it ran `nix develop "$CLONE_DIR" --command direnv allow` —
   against the *sibling's* dev shell, which declares no direnv. Run 32489213400,
   the one run in months whose unfold happened to succeed, reached it and got
   `exec: direnv: not found`, exit 127. direnv is a declared package of *this*
   repo's shell precisely because `build.rs` shells out to it; a sibling's flake
   is not ours to add it to.

### Tests

- `ci/test/sibling-provisioning-test.sh` gains a contract: every job that runs
  cargo against `src/db-backend` provisions `codetracer-native-recorder`. A job
  may satisfy it by delegating to the composite, but only while the composite
  still names the recorder — otherwise dropping the entry there would silently
  un-provision every caller while the suite kept saying "ok".
- `ci/test/unfold-submodules-test.sh` (new, `lint-bash`) drives the real script
  against real repositories with real and nested submodules. It reproduces the
  pre-fix loop's silent partial success first, so the fixture cannot go stale
  without failing.
- `ci/test/direnv-provenance-test.sh` (new, `lint-bash`) fails when a
  `nix develop <ref> ... direnv` names a `<ref>` outside this repository, and
  first checks its own premise — that `ci-base.nix` still declares direnv and
  both shells still compose it — so the rule never points at a shell that
  cannot serve it. Its exact call-site count already caught #659 adding one.

Converting the hand-rolled sibling block also retires three branch-tip pins, so
the ceiling in `sibling-provisioning-test.sh` drops 61 -> 59.


zah:
Superseded by #666, which is the same five fixes linearised onto current `dev`.

This branch became unmergeable once #659 landed: it carried #659's commits via a merge commit, and this repo allows rebase-merge only, so GitHub refused with `This branch can't be rebased`. Rebasing in place would have required a force-push, which the reprobuild pre-push gate correctly refuses as a non-fast-forward. A fresh linear branch was the clean way through.

#666 is merged.

PR #669, opened 2026-08-27
ci: pin the isonim sibling clones to =dev, like every other siblings list

zah:
**One bug, ~15 failing jobs on every pull request.**

`.github/actions/setup-isonim-siblings` was the only `siblings:` list in this repo with **bare** entries. A bare entry asks `clone-siblings` to resolve the revision from the workspace lock for the commit — and these nine repos are not members of codetracer's project in `metacraft-manifests`, so the lock has nothing to say about them:

```
The workspace lock for codetracer@<sha> pins no revision for these
sibling(s): isonim isonim-tui isonim-gpui nim-termctl nim-pty
```

Every other `siblings:` list already carries an explicit `=dev`, and the convention is documented at the top of `.github/workflows/codetracer.yml`:

> A bare entry asks that action to resolve the revision from a *workspace lock* … An entry of the form `name=ref` skips that lookup and clones `ref` directly.

This makes the last list consistent with the other five.

## Why it looked like many unrelated failures

Because everything reaches this action through `Setup dev env`. The casualties included `dev-build`, `nix-build`, `test-non-gui`, `test-ui-tests`, `appimage-build`, `ViewModel headless tests`, `cross-process value-origin`, both `ct-test` gates and `visual-replay-regression-gate` — which reads as a broad breakage rather than one missing `=dev`.

## Not caused by the recent runner routing

Worth stating, since it landed alongside that work: **PR #656 predates the routing entirely and fails the identical set with the same error**, on the *old* `eph-linux-x64` labels. Same jobs, same cause, different runners.

## The better long-term fix, deliberately not done here

Declare these nine in codetracer's project manifest in `metacraft-labs/metacraft-manifests`, so the lock can pin them and the entries can go back to bare — which is what the lock mechanism is *for*, and what the error message itself recommends.

That is a change in another repo, and CI stays red until it lands **and** a new lock is published. This PR unblocks CI now; the manifest change can follow without urgency. Both options are recorded in a comment on the action so the next person does not have to re-derive them.

### DIFF
--- a/.github/actions/setup-isonim-siblings/action.yml
+++ b/.github/actions/setup-isonim-siblings/action.yml
@@ -72,13 +72,41 @@ runs:
       uses: metacraft-labs/metacraft-github-actions/clone-siblings@main
       with:
         gh-token: ${{ inputs.gh-token }}
+        # A BARE entry resolves its revision from the workspace lock. An entry
+        # of the form `name=ref` SKIPS the lock and clones `ref` directly.
+        #
+        # PIN ONLY WHAT THE LOCK CANNOT ANSWER FOR. Every bare entry below is
+        # bare on purpose: the lock pins it, and a lock pin is the reproducible
+        # answer. Adding `=dev` to one of those would silently trade a pinned
+        # revision for a moving branch tip — a real loss of reproducibility,
+        # and invisible, because CI would go on passing.
+        #
+        # The five pinned entries are the ones the lock genuinely cannot
+        # answer: they are not members of codetracer's project in
+        # `metacraft-labs/metacraft-manifests`, so the lock has nothing to say
+        # about them and a bare entry fails the step outright:
+        #
+        #   The workspace lock for codetracer@<sha> pins no revision for these
+        #   sibling(s): isonim isonim-tui isonim-gpui nim-termctl nim-pty
+        #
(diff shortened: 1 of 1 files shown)


zah:
**Corrected — the first revision of this PR contained the exact bug the operator warned about.**

It pinned **all nine** siblings to `=dev`. But the lock only fails to answer for **five** of them (`isonim`, `isonim-tui`, `isonim-gpui`, `nim-termctl`, `nim-pty`) — the error message names those five and says explicitly *"The lock is intact — it answered for the others"*.

So the other four (`nim-everywhere`, `nim-acp`, `nim-agent-harbor`, `nim-agents`) **were** pinned by the lock, and adding `=dev` would have traded four pinned revisions for moving branch tips. Silently: CI would have gone on passing, and nobody would have noticed reproducibility had been given away.

Force-pushed. Now pins exactly the five the lock cannot answer for; the other four stay bare so the lock keeps pinning them. The rule is written into the action so the next person does not repeat it:

> **Pin only what the lock cannot answer for.** Every bare entry is bare on purpose: the lock pins it, and a lock pin is the reproducible answer.

Also updated the PR description's framing of the long-term fix: declare those five in the project manifest and then **delete the `=dev` from all five**, rather than leaving them pinned forever.

PR #674, opened 2026-08-28
ci: provision sibling repos from repro.lock, not the workspace lock

zah:
CI cloned sibling dependencies from a hand-written nine-repo list resolved against the **workspace lock** in `metacraft-manifests`. That lock describes a *repo set* — a human convenience for setting up a workspace, typically a superset of what a build needs — and is not the build's source of truth. `repro.lock` is, and it already has the answer.

> **Base note:** this targets `main`. `dev` was deleted from the remote while this work was in flight and merged into `main`, which is now the mainline (`stable` is the GitHub default but is 17 days behind). Verified rather than assumed.

## Two facts that settle it

**1. The two locks disagree about `isonim` right now.**

```
repro.lock       e82b819d2d05efb6023a966e886fdbec6d1660cb
workspace lock   9936c544925004bada019204b5b111f39d1a54bf
```

CI has been building against a different `isonim` than this repo's own lock pins.

**2. The workspace lock's contents depend on who last pushed.**

`isonim` and the four extras appear in codetracer's lock XML only via `groups="local::isonim"` — they are there because *the pusher's workspace happened to have the IsoNim project checked out*. The same file carries `groups="local::grip"` and `groups="local::reprobuild"` for the same reason. `projects/codetracer.toml` declares only `nim-everywhere`, `nim-acp`, `nim-agent-harbor`, `nim-agents`.

So the previous action's header — *"the project declares them and the lock pins all nine"* — was false, and the mechanism it depended on was incidental to a colleague's local checkout.

## The change

New `.github/actions/provision-repro-lock-siblings` runs the documented consumer path — the same command reprobuild's own CI already runs against this repo:

```sh
nix develop .#ci --command repro develop --all --reset --into="$siblings_root"
```

`repro` comes from codetracer's **own flake**, so there is one pin and no download. Verified end to end: **5 repos at exact SHAs in 6.7 s**, and **no credential needed** — all five are public, so the action takes no `gh-token`.

`setup-isonim-siblings` and its nine-repo list are deleted, including the `=dev` pins added recently — `repro develop` supplies exact SHAs, so pinning to a branch becomes moot. The step moves after `Install Nix` in all eight jobs, since it needs the dev shell; nothing between checkout and Nix used the siblings.

## The four extras are genuinely unnecessary — five independent lines of evidence

`isonim-tui`, `isonim-gpui`, `nim-termctl`, `nim-pty`:

- **`nim-pty`** — zero imports anywhere, in codetracer or any dependency.
- **`nim-termctl`** — zero imports from codetracer; reachable only from `isonim_tui/{input/parser,drivers/*,widgets/image}`, none of which is in the closure of the two isonim-tui modules anything here touches (BFS: 0 edges).
- **`isonim-tui` / `isonim-gpui`** — imported by exactly two files, both in the `vm-collab-integration` lane, which **no workflow runs** and which the justfile records as red anyway.
- **This repo's own script agrees**: `scripts/require-siblings.sh` lists all four as ADVISORY with the sentence *"builds succeed without them"*. Run with the four absent: four warnings, **exit 0**.
- **Two independent declarations omit them**: `repro.lock`'s `depends` and `repro.nim`'s requires. isonim's own `repro.lock` and `.nimble` don't declare them either.

None was declared as a dependency, because none is one.

## Why not `env-flavor: reprobuild`

Read at source. It clones **reprobuild's own** build inputs and explicitly says a consumer's siblings must come from `repro develop --all` *run by the workflow*; it never runs that for the calling repo, and it **does not install Nix**. All eight jobs need Nix and drive everything through `nix develop`, so switching flavor would have replaced their dev env rather than their sibling provisioning.

## Kept

The tailwind placeholder seeding stays. `build-tailwind.sh` *produces* `isonim/build/tailwind-styles.json`, but lanes that compile Nim without `build-once.sh` never run the extractor, and a fresh isonim clone has no `build/`. `staticRead` cannot be caught by its `try/except`, so absence is a compile error.

## Contract tests moved with the change

`sibling-provisioning-test.sh` had an assertion anchored on the deleted block; it now asserts the new source of truth — the action runs `repro develop --all --reset --into=`, hand-writes no sibling set and no 40-hex revision, and `repro.lock` declares exactly five, each with a 40-hex SHA. 26 → 27 assertions.

`nix-provisioning-test.sh` listed `setup-isonim-siblings` as a Nix *provisioner*, which was wrong — it installed no Nix. Replaced with an `is_nix_consuming_step` rule so a future edit hoisting the new action above `Install Nix` fails. **Side finding:** this surfaced that `dev-build` had never been classified as nix-using at all, because its only nix line uses the quoted `- run: "nix develop …"` form the scanner's prefix match cannot see. Documented rather than papered over.

## Two things for elsewhere

- **`setup-reprobuild`'s prebuilt path is currently dead**: release `v0.1.3` publishes **no SHA256SUMS** (so the default `require-checksum` rejects it) and **no darwin-aarch64 archive**. Both misses fall through to building the CLI from source — minutes to hours slower — in eight jobs.
- `request-panel-sidecar-retirement.yml`'s six open-coded `clone-repo@main` calls are untouched: those are recorder repos for a runtime test, not in `repro.lock`, at three different default branches. A `clone-siblings` question, not a `repro.lock` one.


zah:
## This has already landed, byte-for-byte. Recommending closure.

Triaged as part of a sweep of the open PR queue. **The change in this PR is already on `dev`** — it landed as a direct rebase rather than through this PR, so GitHub never closed it.

**Evidence.** Commit [`7c8604c87`](https://github.com/metacraft-labs/codetracer/commit/7c8604c87) on `dev` — *"ci: clone the dependency siblings this repo declares, at the revisions it pins"*, 2026-08-28 — carries the identical 9-file diffstat `+460/-272`. Comparing content rather than offsets: with `index` lines and hunk headers stripped, **every `+`/`-` content line of `gh pr diff 674` is identical to `git show 7c8604c87`**.

- `git merge-base --is-ancestor 7c8604c87 origin/dev` → yes.
- This PR's head `0cbdbdb5` is **not** an ancestor of `dev`, which is exactly why the PR still reads `DIRTY`/`CONFLICTING`.

So the conflict is not work to be redone — it is this branch conflicting with its own landed content plus 266 commits of drift. There is nothing to rebase and nothing to cherry-pick.

**Verdict: superseded.** I am not closing it; that is the maintainer's call. But the `.github/actions/provision-repro-lock-siblings/action.yml` approach, the removal of `setup-isonim-siblings`, the nine converted jobs and the `require-siblings.sh` / `nix/shells/main.nix` / contract-suite updates are all on `dev` today.

PR #694, opened 2026-09-02
docs(cloud): correct prose that describes checks and mechanisms that changed

zah:
A prose sweep of `cloud`, in the pattern already confirmed seven times in these repos: when a check, constant or mechanism changes, the assertion gets corrected and the sentence describing it does not. Every claim here was verified against the thing it describes — the real source, a tree-wide caller grep, a resolved ref — never against another comment.

**Three sentences said work was blocked or expensive when it was not.** `noir_template.nim` said producing the template's opcode count in a browser "needs a new wasm export and a new worker branch". It needs neither: `ci/test/noir-template-acir-count.mjs` already computes exactly that total from the shipping module over the existing ABI, and `web_noir_build.nim` already issues a program-mode compile through the worker. The same paragraph's "exports `nv_compile_vfs` and `nv_test_vfs` and nothing else" and "the only compile a tracer host asks for sets `force_brillig: true`" were both false too. That paragraph carried its own warning that it had gone stale once before, over `nargo test`; it then went stale again in the same shape, so it no longer restates the export or subcommand lists and points at the files that define them.

**Six places described a `nargo info` comparison that was removed** for being against the wrong compiler. The gate compares the ACIR total as an integer against the wasm module; it does not run `nargo info`, does not diff bytes, and does not check the unconstrained counts at all — perturbing one reddens no arm. The gate's own comment beside the comparison already said so; the prose around it did not.

**A user-visible string attributed the count to the wrong compiler**, and two gates asserted only that it contained "nargo info" — certifying a claim about provenance while establishing nothing about the number, and pinning the wrong attribution in place. String and both gates are corrected together.

**A keep/delete verdict rested on a false premise:** the triage doc said `origin/dev` carries neither sibling-pins file. It carries both, byte-identical to cloud's copies. Only `main` lacks them, so the verdict is restated on the premise that survives.

**A deploy header said "`ide.` is the product's one domain"** while the same workflow declares `noirstudio.dev` and `www.noirstudio.dev`. Reading that header as exhaustive is how the verify step came to check two hosts while the deployment declared three — the defect that step was repaired for.

**One correction is load-bearing and left a real gap visible rather than hidden:** `constraints_vm.nim` said `markStale` "is therefore called from the editor's change hook". It has no call site — three occurrences tree-wide, being its definition, one re-export, and that sentence. The pane therefore never labels a stale count, which is the exact state the paragraph above it calls worse than no number. The comment now says that plainly instead of describing an intention as behaviour.

All source changes are comments, docstrings and message text, except the provenance string and the two gate assertions that check it, which move together.

### DIFF
--- a/.github/workflows/deploy-web-codetracer.yml
+++ b/.github/workflows/deploy-web-codetracer.yml
@@ -10,8 +10,16 @@ name: Deploy ide.codetracer.com
 # NAMES (2026-08-29): the hosted CodeTracer web product's domain is
 # https://ide.codetracer.com. An earlier plan put it on `web.codetracer.com` and
 # a later note here described `web.` as a separate authenticated application that
-# had to keep working; both are superseded — `ide.` is the product's one domain,
-# and `web.codetracer.com` is not a name this repo targets any more. Unchanged by
+# had to keep working; both are superseded — `ide.` is the product's PRIMARY
+# domain, and `web.codetracer.com` is not a name this repo targets any more.
+#
+# `ide.` is NOT the only host this deployment serves, and reading this header as
+# if it were is how the verify step came to check two hosts while the deployment
+# declared three. The authoritative list is the descriptor's `origin` plus every
+# `languageOrigins[].origin` — see `CT_WEB_LANGUAGE_ORIGINS` below, which today
+# adds `noirstudio.dev` and `www.noirstudio.dev`. Anything that enumerates hosts
+# (a cache purge, a reachability check) must derive them from that variable and
+# not from this paragraph. Unchanged by
 # that decision: the Cloudflare Pages PROJECT is still named `web-codetracer`,
 # this workflow file is still `deploy-web-codetracer.yml`, and the preview host is
 # still `web-codetracer.pages.dev` — those are infrastructure identifiers, not the
@@ -81,8 +89,10 @@ name: Deploy ide.codetracer.com
 #
 # ONE ENVIRONMENT / ONE PROJECT
 #   The `web-codetracer` Pages project's PRODUCTION branch is `cloud`. Merging
-#   to `cloud` publishes a production deployment served on ide.codetracer.com
-#   (and web-codetracer.pages.dev).
+#   to `cloud` publishes a production deployment served on ide.codetracer.com,
+#   on web-codetracer.pages.dev, and on every host in CT_WEB_LANGUAGE_ORIGINS
+#   (currently noirstudio.dev and www.noirstudio.dev) — one bundle, several
+#   hostnames, each with its own Cloudflare cache.
 #   (Cloudflare decides production-vs-preview by comparing the deployed branch
 #   to the project's configured production branch — see metacraft-labs/web-site
 #   DEPLOY.md.) Create the project once with wrangler before the first run; this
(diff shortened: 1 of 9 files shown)


zah:
**Follow-up filed, so the defect this PR renames does not close with it.**

The `markStale` correction here is load-bearing: it turns a false statement ("is therefore called from the editor's change hook") into an accurate one, but the underlying defect — **the Constraints pane never labels a stale count** — remains open. It is now tracked at **#696**, with the call-site evidence and the open design question (`constraints_vm.nim:27-45`) that explains why it was never simply wired.

Note also that `cloud` landed its own correction of the same sentence while this branch was open, going further than this one (it also fixes `isonim_constraints_view.nim` and records the zero call sites in `ci/test/frontend-reachability-guard.py:337`). This PR's version of that hunk was **dropped** in favour of the one already on the branch rather than reasserting a slightly different count of the same fact; the commit message was amended to say so.

zah:
Landed on `cloud` as **5c5bfe71e5d6a3d30d96656891a442ffd321024f** (all 3 commits, rebased onto `1fa7d045`).

Verified after push: `git ls-remote origin refs/heads/cloud` → `5c5bfe71…`, matching the local SHA.

### CI verdict — the claim was checked, not repeated

The claim given to me was "every failure is in `Setup isonim siblings`". **That wording is wrong**, but the substance holds. Only 11 of 21 failures are literally in `Setup isonim siblings` (first error: `No workspace lock for codetracer (candidates: b4738be9…)`). The rest are in four *other* pre-existing infrastructure families: `Run nix develop … ./ci/lint/bash.sh` and friends failing on `Failed to fetch git repository https://github.com/anza-xyz/crossbeam : fatal: could not read Username`; `Checkout CodeTracer` on a corrupted self-hosted runner workspace; `Run Reprobuild macOS smoke`; and downstream fallout in `Setup rr-backend` and the origin-DAP gate.

**Pre-existence proven independently**, not inferred: run [33671177487](https://github.com/metacraft-labs/codetracer/actions/runs/33671177487) is a `push` on pristine `cloud` at head `b4738be9` — the PR's exact base, containing none of this PR's content. It fails **22** jobs, and this PR's 18 `Codetracer CI` failures are a strict **subset** of them: same job names, same step names, same error text. This PR contributes zero new failures.

The decisive mechanism: `resolve-sibling-rev` logs `candidate SHAs: b4738be9…` — only the **base** commit is consulted, never this PR's head. The lock lives in an external repo (`metacraft-manifests@latest`); nothing in this PR can influence the lookup.

Every failing step is step 4-8 of its job (checkout / dev-env / sibling provisioning). **No failing step runs after any step that reads a file this PR touches** — no job reached Nim compilation. And `deploy-web-codetracer.yml` has no `pull_request` trigger, so that workflow and the three `ci/test/*.sh` scripts it drives are entirely unexercised by PR CI.

Because red CI here proves nothing either way, the Nim changes were verified **locally** instead, on the rebased branch:
- `just test-vm-unit` — **53 files, 1067 cases, 0 failed**
- `just test-vm-unit-js` — **48 files, 991 cases, 0 failed**

Both exactly match the stated green baseline. Also checked: all three changed `.sh` files pass `bash -n`, the workflow YAML parses, and the moved provenance string has no stale matcher left anywhere (`*"nargo info"*` survives only inside comments describing the change).

cspell 9.7.0 (from the nix store — it is not on PATH and `npx` cannot fetch it) over the 11 changed files, differentially against `origin/cloud`: **one** new unknown word, `recompiles` — correctly spelled, simply absent from `project-words.txt`. No misspellings.

Closing in favour of the landed commits.

PR #695, opened 2026-09-02
docs(dev): correct prose that names things the tree does not contain

zah:
A prose sweep of `dev`, in the pattern confirmed repeatedly in these repos: when a check, constant or mechanism changes, the assertion gets corrected and the sentence describing it does not. Every claim was verified against the thing it describes — the real source, a tree-wide caller grep, the cited command actually run — never against another comment.

### Citations to things that do not exist
- `language_detection.nim` credited a `lang_detection_test.nim` for asserting `detectLangFromPath`'s return value. No such file. The test that does make that assertion is `src/tests/cli/lang_enum_contract_test.nim`, so the citation is re-pointed rather than deleted — the coverage is real.
- `webcrypto_verifier.nim` credited a `test_webcrypto_verifier.nim` under the heading "This is exercised, not merely written". No such file either; the coverage is real via `ci/test/identity-webcrypto.sh`, wired at `justfile:2993`. Also re-pointed.
- `cow_namespace_crossread_test.rs`'s M4 generator claimed a Nim test read its fixture back, "closing the bidirectional wire-format loop the M5/M6 Rust replay-time write path depends on". There is no such consumer — no `*.nim` file reads a `.cowbt` at all. **The loop is open.** The comment now says so and gives the shape of the missing test instead of claiming a guarantee nothing provides, and its green line no longer announces a cross-read that never happens.
- `release_gate.nim`'s ordinal-drift example cites `src/tui/src/lang.rs`, deleted in `84a8b633`.

### Procs described as called, with no callers
`isonim_app.nim` said two procs are "called from ui_js.nim" — zero call sites tree-wide, and `ui_js.nim` says the opposite ("The standalone isonim_app shell remains disabled here"). `search_results.nim` claimed three hooks are called from `onSearchResultsUpdated`; `syncSearchResultsClearHook` is declared, assigned, and never called.

### Numbers stale next to the command that refutes them
`justfile:2555` said "76 of the 220 files (`test_lane_all_files | wc -l` for the 220)". Running it gives **274**. The reach budget said "sixteen of the nineteen"; `budget_for()` sums to 20. `require-siblings.sh` said five trace-format deps (7) and 223 import sites (532 lines across 226 files). Rather than writing fresh numbers that will rot the same way, these now name the command that computes them.

### Green paths and instructions that do not work
`just build-tailwind` does not exist (the step is `vm-test-prereqs`). `test-rust-flow-all` prints "All Rust flow tests passed!" while running only the stable arm — why the other two are skipped is recorded nowhere, and the comment now says that rather than inventing a reason. The identity mutation gate's `RESULT: OK` claimed every assertion family has an arm that reddens it; it never quantified over families.

### A real defect found behind reassuring prose
Two viewmodels documented `ct/history-jump` as handled by "the existing backend dispatcher". It parses `Location`, which is `#[serde(default)]`; neither payload shares a field name with it, so every field defaults and `location_jump` runs on a zeroed `Location` — **a jump to step 0 that returns success**. Marked as a known defect in both viewmodels; no behaviour change here, since the fix is to flatten the payloads or give the command its own argument type.

`dap_dialect.md`'s `task.rs` line citations had all drifted; the numbers are dropped in favour of symbol names, which do not rot.

### DIFF
--- a/ci/lib/test-lane-files.sh
+++ b/ci/lib/test-lane-files.sh
@@ -346,16 +346,26 @@ test_lane_files() {
 		;;
 
 	host-instantiations)
-		# NOT test files. The platform facade's two host instantiations —
-		# production modules — compiled on the backend they actually ship on.
+		# NOT test files. Production modules of the platform facade, compiled
+		# on the backend they actually ship on. Four are echoed at the bottom
+		# of this arm; read that list, not a count written here.
 		#
-		# WHY THIS LANE EXISTS, and it is not a hypothetical. Both of these
-		# modules are `{.error.}` on the C backend by design, so `vm-unit`
-		# cannot see them; and neither is imported by any suite in `vm-unit-js`,
-		# because the whole point of `platform/web_platform.nim` is that it
-		# reaches no browser API and can therefore be tested without these.
-		# The result was a hole exactly the shape of these two files: NOTHING
-		# IN CI COMPILED THEM AT ALL.
+		# WHY THIS LANE EXISTS, and it is not a hypothetical. The argument is
+		# about the two host INSTANTIATIONS — `desktop_electron.nim` and
+		# `web_browser.nim`. Both are `{.error.}` on the C backend by design,
+		# so `vm-unit` cannot see them; and neither is imported by any suite in
+		# `vm-unit-js`, because the whole point of
+		# `platform/web_platform.nim` is that it reaches no browser API and can
+		# therefore be tested without these. The result was a hole exactly the
+		# shape of those two files: NOTHING IN CI COMPILED THEM AT ALL.
+		#
+		# The other two in the list are here for their own reasons, not this
+		# one. `platform_host.nim` is here for its Electron arm (spelled out
+		# below). `opfs_volume.nim` is the web instantiation's OPFS volume: it
+		# is also `{.error.}` on C, but unlike the two above it IS driven by a
+		# suite — `test_opfs_volume.nim`, added to `vm-unit-js` against a fake
+		# `navigator.storage` — so for it this lane is a second compile rather
+		# than the only one.
 		#
 		# It was not theoretical for long. `web_browser.nim` reached `dev` at
 		# ed9d6021 with a doc comment after the closing paren of an object
--- a/ci/test/identity-token-mutation.sh
+++ b/ci/test/identity-token-mutation.sh
@@ -6,13 +6,27 @@
 # `src/frontend/viewmodel/identity/token.nim` decides whether a signed identity
 # token is accepted, which band of its life it is in, and whether a subject is
 # revoked. A suite that reports ten green cases over it is worth exactly what
-# the evidence says it is, so this file supplies the evidence: one mutation per
+# the evidence says it is, so this file supplies the evidence: a mutation per
 # assertion family, each verified to redden THE CASE WRITTEN FOR IT.
 #
 # A mutation caught by some other case is a MISS, not a kill. Every arm below
 # names the case it expects to go red, and several additionally name the cases
 # that must stay GREEN — because an arm that reddens everything proves only
 # that the suite noticed a change, not that the assertion in question can fail.
+#
+# WHAT THIS FILE DOES NOT DO, and the PASSED line now says so too: it never
+# reads the suites' case NAMES. The arms are a hand-written list, each matched
+# against one name, so "every declared arm killed the case it names" is the only
+# universal this gate can support. A case that no arm targets is simply absent
+# from the run, and several cases in the three suites are in that position today.
+#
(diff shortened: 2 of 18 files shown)


zah:
**Follow-ups filed, so the two defects this PR renames do not close with it.**

- **#697** — the Rust `.cowbt` on-disk format has never been read back by Nim. This PR corrects the M4 generator comment that claimed the loop was closed; the loop is still open. Evidence: `src/db-backend/tests/cow_namespace_crossread_test.rs:176-184`, and a committed 24576-byte fixture (`tests/fixtures/cow_namespace/cow_btree_rust_typea.cowbt`) that nothing reads.
- **#698** — `ct/history-jump` silently jumps to step 0 and reports success. Neither caller's payload shares a field name with `Location`, which is `#[serde(default)]` at the container level (`task.rs:503-505`), so every field defaults and `location_jump` reads only `rr_ticks`. Verified end to end through `dap_server.rs:2110` → `dap_handler.rs:2994`.

**Commit hygiene:** `9c6dde00` mixed two justfile hunks into a task.rs-only commit whose message did not mention them. They were **not** another agent's work — they are this sweep's own `test-renderer-browser` and `test-rust-flow-all` corrections, described in `0f4cbff5`'s message but missing from its diff. The two commits were rebuilt so each message matches its contents; the final tree is byte-identical to what was pushed here.

zah:
Landed on `dev` as **d05e45f97e449c0313ffbdf9bdd29d77de911a11** (13 commits, rebased onto `842d9927`).

Verified after push: `git ls-remote origin refs/heads/dev` → `d05e45f9…`, matching the local SHA.

### `9c6dde00` — split, and the diagnosis was wrong

It was reported that `git add -A` in a shared checkout swept **another agent's** in-progress justfile work into this commit. **That is not what happened, and the evidence is decisive.** The two justfile hunks are *this sweep's own* corrections: `test-renderer-browser`'s `just build-tailwind` → `vm-test-prereqs`, and `test-rust-flow-all`'s "All Rust flow tests passed!". Both are described in **`0f4cbff5`'s own commit message** — which claimed three justfile corrections but contained only one — and both are listed in this PR's description. They were committed one commit early, not stolen.

Corroborating: the shared checkout has no uncommitted justfile work (only three unrelated `non-nix-build/windows/` files), and no other branch in the repo carries these hunks.

So they were **split, not dropped** — dropping would have discarded correct, verified work of this sweep's own. The two commits were rebuilt as:
- `3fa47891` — `src/db-backend/src/task.rs` **only**, keeping `9c6dde00`'s message, which now describes exactly its contents.
- `03626bcd` — `justfile` **only**, carrying `0f4cbff5`'s message, which now describes exactly its contents, plus a note recording the mis-commit.

The reconstructed tree is **byte-identical** to the original branch tip (`git diff origin/sweep/prose-dev HEAD` empty before rebase). Both justfile claims were re-verified against the tree: `build-tailwind` is not a recipe (only `scripts/build-tailwind.sh`, invoked by `vm-test-prereqs` at `justfile:2459`), and `test-rust-flow-all` invokes only `test-rust-flow-stable` while `test-rust-flow` and `test-rust-flow-nightly` exist at lines 1451 and 1471.

### Verification run

- `just test-vm-unit` — **34 files, 763 cases, 0 failed**; `just test-vm-unit-js` — **30 files, 696 cases, 0 failed**. (A transient nix `narHash` mismatch on the third-party `spectrum-os` flake input appeared during direnv setup; direnv recovered and both lanes ran to completion.)
- All 18 changed files audited for non-comment changes: every `.nim`, `.rs` and `.mjs` file is **comment-only**. The only logic-adjacent edits are output strings in `identity-token-mutation.sh` (the `arms` counter it now prints is defined at line 100), `require-siblings.sh` descriptions, and justfile echo text. All five changed shell scripts pass `bash -n`; the justfile parses.
- Spot-checked the corrections themselves rather than trusting them: `lang_detection_test.nim` and `test_webcrypto_verifier.nim` genuinely do not exist while `lang_enum_contract_test.nim` really does assert `detectLangFromPath` (20 references); `createIsoNimApp`/`mountIsoNimApp` have zero call sites outside their own file; `syncSearchResultsClearHook` is declared and assigned but never invoked; `src/tui/src/lang.rs` really was deleted in `84a8b633`; all five re-pointed `task.rs` symbols resolve; `budget_for()` really sums to 20 with 16 webexcluded (so the stale "nineteen" was wrong and the "sixteen" was right); and the corrected `nimLanguage.test.mjs` run instruction works as written — **37 passed, 0 failed**, no node flags.
- cspell 9.7.0 differential vs `origin/dev`: two new unknown words, `deserialisation` and `deserialises` — correctly spelled and consistent with `serialised`/`serialising`/`serialises` already in `project-words.txt`, just not added. No misspellings.

Closing in favour of the landed commits.

PR #700, opened 2026-09-03
ci: obtain aws via nix shell, and mount the dmg instead of unarchiving it

zah:
Two red jobs in run `33742876598`, each failing for a reason with nothing to do with the thing the job checks.

### `appimage-cross-distro` — all five legs, exit 127
Jobs `100625123843` (debian:12), `100625123865` (fedora:40), `100625123866` (ubuntu:22.04), `100625123986` (ubuntu:24.04), `100625124039` (archlinux:latest). Step **Download AppImage**:

```
aws: command not found
##[error]Process completed with exit code 127.
```

The `Install AWS CLI` step above it **succeeded** in every leg. `nix profile install` writes to `$HOME/.nix-profile`, whose `bin/` reaches PATH only via nix's profile snippet; each `run:` step is a fresh non-login, non-interactive shell (`/usr/bin/bash -e {0}`) that sources no profile. The binary was on disk and unreachable by name.

The previous fix in this area ungated the install step so it would actually run. That was correct and insufficient — it addressed why the step was *skipped*, not why the *name does not resolve*, and the symptom survived it unchanged. Now uses `nix shell nixpkgs#awscli2 --command aws`, the form `appimage-lib-check` and `dmg-lib-check` already use, both of which got past their download in the same run.

### `dmg-lib-check` — exit 126
Job `100624879771`, step **Check if ct starts**:

```
./CodeTracer/CodeTracer.app/Contents/MacOS/bin/ct: Permission denied
##[error]Process completed with exit code 126.
```

126 is "found but not executable" — not a crash of `ct`. `7z` reported `Everything is Ok` and unpacked all 32177 files first. It reads the DMG's HFS+ image as an archive and does not restore POSIX mode bits, so the check could not pass whatever the DMG contained.

Replaced with `hdiutil attach`, which mounts the volume with its real permissions, so the executable bit tested is the one the artifact actually ships. A `chmod +x` would also have turned the job green — and would have forged the exact property this job exists to assert, letting a genuinely non-executable `ct` ship unnoticed.

### Why now
Neither was visible until today: `dmg-build` had not reached a runner in ~30 runs, and `appimage-build` depends on lints that were red until the shellcheck backlog landed. The steps are old; the exposure is new.

### Verification
- `.github/workflows/codetracer.yml` parses; both jobs' step lists confirmed.
- Static gates re-run locally under bash 5.2: `nix-provisioning-test.sh` 2/2, `windows-runner-bootstrap-test.sh` 66/66, `required-jobs-test.sh` all contracts hold.
- `sibling-provisioning-test.sh` fails 2 of 26 — **pre-existing at `dev` tip `1a427e3e`, unrelated to this change** (branch-tip sibling ceiling; see below).

### Note for reviewers
This PR is also being used to obtain a `Codetracer CI` run that survives to completion, since pushes to `dev` are superseded and the last 100 runs contain zero successes.

### DIFF
--- a/.github/workflows/codetracer.yml
+++ b/.github/workflows/codetracer.yml
@@ -3854,11 +3854,58 @@ jobs:
           nix shell nixpkgs#gnupg --command gpg --verify ./CodeTracer.dmg.asc ./CodeTracer.dmg
 
       - name: Extract dmg
+        # WHY `hdiutil` AND NOT `7z x`.
+        #
+        # This step used to run `nix shell nixpkgs#p7zip --command 7z x
+        # ./CodeTracer.dmg`, and the "Check if ct starts" step below then died:
+        #
+        #     .../<step>.sh: line 1: ./CodeTracer/CodeTracer.app/Contents/MacOS/bin/ct: Permission denied
+        #     ##[error]Process completed with exit code 126.
+        #
+        # Run 33742876598, job 100624879771, on m3-mcl-005. Exit 126 is
+        # "found but not executable", not "not found" and not a crash of `ct`:
+        # 7z reported `Everything is Ok` and unpacked all 32177 files first.
+        #
+        # 7z reads the DMG's HFS+ image as an archive and does not restore
+        # POSIX mode bits, so every extracted file lands non-executable. That
+        # makes the check unable to pass no matter what the DMG contains, which
+        # is why this is an instrument defect and not a packaging one.
+        #
+        # It went unseen because it was unreachable: `dmg-lib-check` needs
+        # `dmg-build`, and 33742876598 is the first run in ~30 where dmg-build
+        # reached a runner at all. The step is as old as the job; the exposure
+        # is new.
+        #
+        # `hdiutil attach` mounts the real HFS+ volume with its permissions
+        # intact, so the executable bit is the one the DMG actually ships. That
+        # matters beyond making the check pass: `chmod +x` after 7z would also
+        # turn the job green, but it would forge the very property this job
+        # exists to assert, and a genuinely non-executable `ct` in a shipped
+        # DMG would then pass unnoticed. This way the check still fails if the
+        # artifact is really broken.
+        #
+        # `-nobrowse` keeps the volume out of Finder; `-readonly` avoids any
+        # write-back to the artifact we just gpg-verified. The contents are
+        # copied out to ./CodeTracer so the paths every later step uses are
+        # unchanged, and the volume is detached before those steps run so
+        # nothing holds a mount on this persistent runner.
         env:
           AWS_ACCESS_KEY_ID: ${{ secrets.R2_CODETRACER_BUCKET_ACCESS_KEY_ID }}
           AWS_SECRET_ACCESS_KEY: ${{ secrets.R2_CODETRACER_BUCKET_ACCESS_KEY }}
         run: |
-          nix shell nixpkgs#p7zip --command 7z x ./CodeTracer.dmg
+          set -euo pipefail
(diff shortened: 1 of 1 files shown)


zah:
## Both halves of this are already on `dev`, including the comment prose. Recommending closure.

Triaged as part of a sweep of the open PR queue.

**1. `hdiutil attach` instead of unarchiving the dmg — landed as [`bf7bfd54d`](https://github.com/metacraft-labs/codetracer/commit/bf7bfd54d)** *("ci: mount the dmg instead of unarchiving it, and get hdiutil from /usr/bin", Sep 4).* `origin/dev:.github/workflows/codetracer.yml` line 4558 already reads

```
hdiutil attach ./CodeTracer.dmg -mountpoint "$mountpoint" -nobrowse -readonly
```

together with the `export PATH="$PATH:/usr/bin"`, the `trap ... hdiutil detach`, the `cp -R "$mountpoint"/*.app` and the `test -x .../bin/ct` guard. No `7z x ./CodeTracer.dmg` remains anywhere on `dev`.

**2. `aws` via `nix shell` in appimage-cross-distro — landed as [`146237152`](https://github.com/metacraft-labs/codetracer/commit/146237152)** *("ci(appimage): get `aws` from `nix shell` in the cross-distro download step", Sep 3).* `dev` line 4784 already reads `nix shell nixpkgs#awscli2 --command aws --endpoint-url=... s3 cp ...`; the `Install AWS CLI` / `nix profile install nixpkgs#awscli2` step is gone, and dev line 4730 even carries the note `# This step (and a since-removed "Install AWS CLI" step)...`.

Neither commit is an ancestor of this PR's head `2e5693dc` — they were applied to `dev` independently. `gh pr diff 700` still shows `+105/-14` only because it diffs against merge-base `87f20a4f`; the head is 2 ahead and **237 behind** `dev`. Merging would be a no-op at best.

### The red CI here is stale-base, not this change — and it is worth recording why

I checked the job logs directly (`repos/.../actions/jobs/<id>/logs`, identity confirmed via `Complete job name:` in each) rather than trusting the summary:

- **`appimage-cross-distro (debian:12)`** (job `100773536953`) — *this PR's own change succeeded*: `download: s3://codetracer/CodeTracer-700/merge-amd64.AppImage to ./CodeTracer.AppImage`, 325.6 MiB. The job then died downstream in `test-appimage-cross-distro.sh` with ``ct_unwrapped: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.38' not found``.
- **`appimage-lib-check`** (job `100773536839`) — same `GLIBC_2.38 not found`.
- **`dmg-lib-check`** (job `100773984876`) — *`hdiutil attach` succeeded* (`/dev/disk17s2 Apple_HFS /private/var/.../codetracer-dmg-iPpIO9`), the `test -x` guard passed, and the job died in the **next** step: `dyld[66390]: Library not loaded: @executable_path/../Frameworks/libcrypto.3.dylib`, exit 134. That is the genuine packaging defect this PR was designed to stop masking — 7z's lost mode bits previously surfaced it as exit 126.
- **`nix-build`** (job `100770233095`) — unrelated `db-backend` Rust unit tests: `ctfs_trace_reader::tests::combined_mcr_and_steps_opens_via_materialized_reader` and `..::step_stream_read_by_structural_presence_when_step_bit_clear`; `test result: FAILED. 630 passed; 2 failed`.

All four are green on `dev` today (runs `34203242916`, `34184025734`, `34169092836`: `nix-build`, `appimage-lib-check`, `dmg-lib-check` and all five `appimage-cross-distro` legs `success`).

**Verdict: superseded.** Not closing it — maintainer's call — but the work is on `dev` as `bf7bfd54d` + `146237152`.

Issue #152, opened 2025-05-30
Comments are grayed out similar to unvisited lines of code


Franz-Fischbach:
While inside of a function, comments get grayed out similar to unvisited lines of code where as other comments that are not inside of the current function are not grayed ot.

![Image](https://github.com/user-attachments/assets/4ce1cf42-9017-4c7c-9d61-c9c8baf08e20)

Steps to reproduce:

1. Open CodeTracer using ` ct run ./examples/noir_space_ship`
2. Navigate to the iterate_asteroids function from the call trace

Expected result:
All comments should be in the same color

Actual result:
Comments inside the current function are treated as unvisited lines of code and are grayed out

PR #701, opened 2026-09-03
ci: resolve the six IsoNim-family siblings from the workspace lock

zah:
Unblocks `ci-verdict` **step 5** (`sibling-provisioning-test.sh`), which has been failing on every run and taking steps 6-14 with it — including step 14, the verdict itself. The gate built to distinguish "tests failed" from "tests never ran" has never reached the step that says which.

Paired with metacraft-labs/metacraft-manifests@8ddec34f. **Both halves are required**; either alone changes nothing observable.

## The gap, and why the obvious fix had already been tried and failed

`clone-siblings` resolves a bare `siblings:` entry from the per-commit workspace lock. Six repos the IsoNim provisioning needs were not in codetracer's project in `metacraft-manifests`, so the lock could not answer for them and a bare entry failed `Setup dev env` outright. The workaround was `=dev` — a branch-tip pin, which `sibling-provisioning-test.sh` counts against a shrink-only ceiling. The guard therefore forbade the only workaround for a gap the manifest created.

An earlier manifests commit had **already declared** `isonim` and `runquota`, and nothing changed. That is the part worth recording, because it looked like the fix had landed:

> **Declaring a repo does not put it in a lock.** `publish-workspace-lock.yml` re-anchors each mainline commit's lock from its *parent*, copying the repo **set** verbatim. A repo declared after a lineage began never enters it. The lock published for `dev` HEAD `79275598` *after* that declaration still named neither repo.

`metacraft-manifests@8ddec34f` does both halves: it declares the four still-missing repos (`isonim-tui`, `isonim-gpui`, `nim-termctl`, `nim-pty`) and **seeds the `79275598` lock** with all six. The re-anchor carry propagates them from there.

## Scope correction

The gap is **six** repos, not two, and the `=dev` overrides are **eight** entries, not two. Fixing only `isonim`+`runquota` would have left four overrides with no lock to resolve from and `sibling-provisioning-test.sh` failing on exactly the assertion it fails on now.

## Revisions

Each is the repo's `dev` tip as of 2026-09-03 — precisely what the `=dev` overrides already resolved to at that moment. Pinning **freezes the revision CI was already building against** rather than choosing a different one: the change is reproducibility, not content.

## Verified locally (bash 5.2 + GNU sed, CI-equivalent layout)

| suite | before | after |
|---|---|---|
| `ci/test/sibling-provisioning-test.sh` | 2 of 26 failed | **all 26 passed** |
| `ci/test/build-once-workspace-test.sh` | 4/4, 10 gaps | **4/4, 8 gaps** |

The pass is **not vacuous**: both runs report identical populations — 95 sibling entries across 16 blocks, 9 entries in `setup-isonim-siblings`, the same expected IsoNim set, the same 26 assertions. Only the two failing assertions flipped.

Branch-tip entries go **67 → 59** against `BRANCH_TIP_CEILING=59`. The ceiling is **left at 59, not lowered** — 59 entries genuinely remain, so lowering it would be false. That `67-8=59` lands exactly on the ceiling is not a coincidence.

`ci-verdict` steps 2-9 and 11 all pass locally under a CI-equivalent environment (step 10 fails only here, identically on pristine `dev`: its fake GitHub API will not start under this sandbox).

## Not verified

**Whether `ci-verdict` reaches step 14 in CI.** It `needs:` 27 jobs including the Windows lanes, which queue ~15h, so the observation is 12-18h out. This PR is the run that produces it; it is not inferred here.

## Follow-up, not done here

All 59 remaining branch-tip entries name repos the published lock **already pins**. They are convertible by deleting `=dev` alone, with no manifests change. That is the next decrement of the ceiling.

### DIFF
--- a/.github/actions/setup-isonim-siblings/action.yml
+++ b/.github/actions/setup-isonim-siblings/action.yml
@@ -81,10 +81,11 @@ runs:
         # revision for a moving branch tip — a real loss of reproducibility,
         # and invisible, because CI would go on passing.
         #
-        # The five pinned entries are the ones the lock genuinely cannot
-        # answer: they are not members of codetracer's project in
-        # `metacraft-labs/metacraft-manifests`, so the lock has nothing to say
-        # about them and a bare entry fails the step outright:
+        # ALL NINE ARE BARE, and that is new. Five of them — isonim,
+        # isonim-tui, isonim-gpui, nim-termctl, nim-pty — carried `=dev` until
+        # metacraft-manifests@8ddec34f, because they were not members of
+        # codetracer's project there, so the lock had nothing to say about them
+        # and a bare entry failed the step outright:
         #
         #   The workspace lock for codetracer@<sha> pins no revision for these
         #   sibling(s): isonim isonim-tui isonim-gpui nim-termctl nim-pty
@@ -94,18 +95,23 @@ runs:
         # ct-test gates — because they all reach this action through
         # `Setup dev env`.
         #
-        # THE PROPER FIX is to declare these five in codetracer's project
-        # manifest so the lock pins them too, and then delete the `=dev` from
-        # all five. It is a change in `metacraft-manifests` and CI stays red
-        # until it lands and a new lock publishes, which is the only reason it
-        # is not done here. Until then these five, and ONLY these five, are
-        # unpinned.
+        # TWO changes were needed, and only the first had been made. The
+        # project declaration (3c625e7) covered isonim and runquota; the other
+        # four stayed undeclared, and — the part that made the first look
+        # ineffective — declaring a repo does not put it in a lock.
+        # `publish-workspace-lock.yml` re-anchors each mainline commit's lock
+        # from its PARENT, copying the repo SET verbatim, so the 99-repo
+        # lineage in flight simply never grew the new members. The lock for dev
+        # HEAD 79275598, published after that declaration, still pinned none of
+        # them. metacraft-manifests@8ddec34f declares the remaining four AND
+        # seeds that lock with all six, so the carry propagates them from here
+        # on.
         siblings: |
-          isonim=dev
-          isonim-tui=dev
-          isonim-gpui=dev
-          nim-termctl=dev
-          nim-pty=dev
+          isonim
+          isonim-tui
+          isonim-gpui
+          nim-termctl
+          nim-pty
           nim-everywhere
           nim-acp
           nim-agent-harbor
--- a/.github/workflows/codetracer.yml
+++ b/.github/workflows/codetracer.yml
@@ -4722,7 +4722,10 @@ jobs:
             nim-stackable-hooks=dev
             codetracer-visual-replay=dev
             codetracer-native-backend=dev
-            runquota=dev
+            # Bare since metacraft-manifests@8ddec34f: runquota is a member of
+            # codetracer's project there and the lock pins it, so the revision
+            # comes from the lock rather than from whatever `dev` points at.
+            runquota
 
       - name: Stage codetracer-trace-format into libs/
         # The nix flake build below overrides the codetracer-trace-format input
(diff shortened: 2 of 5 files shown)


zah:
## Superseded as written — 4 of the 5 files are already at or past their intended end state. ~3 lines of live residue remain.

Triaged as part of a sweep of the open PR queue. Two things came out of it: the premise is weaker than the title suggests, and most of the change has already landed by other routes.

### The workspace lock was never the gate for an all-explicit list

`clone-siblings.sh` lives in `metacraft-labs/metacraft-github-actions` (`clone-siblings/clone-siblings.sh`, branch `dev`), reached here through `setup-dev-env@dev` from 15 call sites. `NEED_LOCK` is set in exactly one place —

```sh
for entry in "${SIBLINGS[@]}"; do
  ...
  [ -z "${rf}" ] && NEED_LOCK=1     # rf = text after '='; empty only for a BARE entry
done
```

— and consumed in exactly one place, `lock_unavailable()`, which `exit 1`s **iff** `NEED_LOCK -eq 1` and otherwise emits `::warning::`. So a missing manifests repo or an unpublished lock is fatal only when at least one entry is bare.

The other three gates cannot fire on an all-explicit list either: the `MISSING` hard error is inside `if [ -z "${ref}" ]`; the per-sibling resolver-defect exit 5/6 is guarded by the same test; and the lock-override check is `error` only under `on-lock-override: error`, whose default is `warn` (`case "${ON_LOCK_OVERRIDE:=warn}"`) — and `codetracer` sets `on-lock-override` **nowhere**.

A 40-hex SHA is additionally the *quiet* shape: `is_sha40` suppresses the "not reproducible" warning a branch-tip ref would emit, and `ci/test/sibling-provisioning-test.sh` classifies it as `sha`, exempt from `BRANCH_TIP_CEILING=56`. Pinning is silent on all four channels.

This is not theoretical — it is the mechanism `dev` relies on today (see the table below).

### Per-file status on `dev` (`8e93d9f28`)

| file | state |
|---|---|
| `.github/actions/setup-isonim-siblings/action.yml` | **Deleted** by `7c8604c87` (PR #674's content, landed by rebase). Modify/delete conflict; the hunk is dead code. |
| `ci/test/build-once-workspace-test.sh` | Both `KNOWN_GAPS` lines **already removed** (defect #152, via `c6eb6a12b`/`c7c44efeb`), with differently-worded replacement prose. |
| `.github/workflows/launcher-recorder-e2e.yml` | Both names **already present** (lines 601–602) — but as `isonim=9936c5449...` / `runquota=f1ca742d1...`, pinned to `flake.lock`'s revs, not bare. This is the mechanism above, in use. |
| `ci/verdict/workspace-lock-freshness.sh` | Rewritten far more thoroughly by `626265447` (+384/-45), replacing the hard-coded `PROBE_SIBLING="codetracer-trace-format"` with a *derived* set of declared bare names. This PR's change here is comment-only and strictly weaker. |
| `.github/workflows/codetracer.yml` | **Not landed.** `runquota=dev` (5629), `isonim=dev` (6517), `runquota=dev` (6533) still present, byte-identical context. |

The external precondition does hold and has propagated: the manifests lock pins all six of `isonim`, `runquota`, `isonim-tui`, `isonim-gpui`, `nim-termctl`, `nim-pty` for `79275598`, for `8e93d9f28` and for current `dev`.

### What is worth salvaging, and the reason I did not just do it

The live residue is three lines in `.github/workflows/codetracer.yml` (5629, 6517, 6533): `runquota=dev` → `runquota`, `isonim=dev` → `isonim`, `runquota=dev` → `runquota`.

I stopped short of making that change, for a reason that should be recorded rather than rediscovered. **The premise and the on-`dev` SHA pins are in real tension.** The comment beside the `launcher-recorder-e2e.yml` pins argues the SHAs must equal `flake.lock`'s revs so that the tup driver (which reads `../isonim` by relative path) and the nix driver (which reads `flake.lock`) see one tree — citing run `33021054045`, where a mismatch broke the `ci` shell. Going bare re-points those checkouts at whatever the *workspace lock* pins, which is not guaranteed to equal `flake.lock`. **Today they already disagree**: the lock pins `isonim 2a24d954…` while the workflow pins `isonim 9936c544…`.

Converting `=dev` to bare is therefore a behaviour change (branch tip → lock-pinned rev), not a cleanup, and it needs a CI run to validate. The runner pool is not currently servicing jobs — `queue-watchdog` is reporting `CI IS NOT BEING SERVICED ... 10 queued, 0 running` — so I could not have validated it, and pushing it blind is how the run-`33021054045` failure comes back.

**Verdict: superseded as written.** Not closing it — maintainer's call. Suggested disposition: close this, and open a 3-line PR against current `dev` for the `codetracer.yml` entries alone, with a CI run behind it. Keep the `launcher-recorder-e2e.yml` half separate until the lock/`flake.lock` divergence above is resolved, since that is the half that can break the `ci` shell.

PR #703, opened 2026-09-03
fix(appimage): ship glibc and its loader, and start binaries through it

zah:
## The defect

The AppImage started on **one of the five distros we test** — rolling-release Arch.

| distro | glibc | before |
|---|---|---|
| `archlinux:latest` | ~2.42 | pass |
| `fedora:40` | 2.39 | fail |
| `ubuntu:24.04` LTS | 2.39 | fail |
| `debian:12` stable | 2.36 | fail |
| `ubuntu:22.04` LTS | 2.35 | fail |

The cross-distro job only began producing verdicts today; until then it died at `aws: command not found` on every leg, which is why this shipped unseen.

## Were the bundled libraries being loaded?

Yes. Checked on all four failing legs, every `required by` path is inside the extracted AppDir:

```
ct_unwrapped: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_ABI_GNU2_TLS' not found
    (required by /tmp/appimage_extracted_.../bin/../lib/libbpf.so.1)
```

The bundled libs load, then ask the **host** libc for a symbol version it cannot supply. glibc was excluded from the bundle deliberately (`lddtree -l ... | grep -v glibc`) and `PT_INTERP` pointed at the host's loader, so libc and the loader were the only two things still coming from the host.

ubuntu:22.04 and debian:12 also show a line whose requirer is **`ct_unwrapped` itself**, not a bundled dependency:

```
ct_unwrapped: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.38' not found
    (required by /tmp/appimage_extracted_.../bin/ct_unwrapped)
```

so this was never fixable by re-linking one library.

## One root cause, two symptoms

We build against a glibc newer than the oldest host we ship to. The toolchain glibc is **2.40** (`glibc-2.40-224`, from the `codetracer-toolchains` pin).

- `GLIBC_2.36` / `GLIBC_2.38` are ordinary **version floors** — they vanish on the 2.39 hosts.
- `GLIBC_ABI_GNU2_TLS` is **not a version**, it is an ABI marker on objects using GNU2/TLSDESC thread-local storage. It is the only error left on ubuntu:24.04 and fedora:40, and it is why Arch passed and nothing else did. **This is the binding constraint**; no version floor explains it.

**Checked against the store path rather than inferred** (an earlier revision of this PR asserted 2.42 and was wrong — see 85f3c0ab). `glibc-2.40-224`'s `libc.so.6` defines version sets up to `GLIBC_2.39` **and** defines `GLIBC_ABI_GNU2_TLS`, so the marker arrived in 2.40. That makes the fix checkable without CI: the union of every requirement across the four failing legs is `GLIBC_2.36`, `GLIBC_2.38`, `GLIBC_ABI_GNU2_TLS`, and the bundled 2.40 provides all three.

## The fix

Ship glibc and its loader, and start binaries through it:

```
$APPDIR/bin/ld-linux-x86-64.so.2 --library-path $APPDIR/lib $APPDIR/bin/<name>.real
```

`PT_INTERP` must be absolute and the kernel resolves it before our code runs, so neither `$ORIGIN` nor `LD_LIBRARY_PATH` can redirect it — and an AppImage extracts to an unpredictable directory. Invoking the loader explicitly is the only portable option.

The alternative — rebuild against an older glibc — needs a floor nobody has picked yet and would mean overriding the toolchain glibc that every CodeTracer repo shares through the `codetracer-toolchains` pin.

**Trade-off:** the artefact grows by the glibc closure, and is now coupled to the loader it ships. `--library-path` is a loader *argument*, not `LD_LIBRARY_PATH`, so it is not inherited — Electron stays deliberately unwrapped on the host loader and our glibc cannot leak into it.

### Why the loader lives in `bin/`, not `lib/`

Starting a program as `ld.so prog` makes `/proc/self/exe` name the **loader** — measured, and `--argv0` fixes `argv[0]` only. Nim's `getAppDir()` reads `/proc/self/exe`, and this tree uses it in a dozen places to locate sibling binaries, `share/` and the prefix. Keeping the loader beside the programs means `getAppDir()` still returns `<AppDir>/bin`, so **every one of those call sites keeps working untouched**.

Only the four callers wanting the *full path to the executable* need help; they get it from `CODETRACER_APP_FILENAME` via the new `paths.ctAppFilename()`. One of them is `installNativeBpf`, which `setcap`s "the ct binary" — left alone it would have granted capabilities to the bundled dynamic loader.

## Verification

A passing `--version` is not evidence of portability: it also passes when the host glibc merely happens to be new enough. The smoke test now asserts the **resolved libc path**, via `ld.so --list` and again via `LD_DEBUG=libs` on the real process, failing the distro unless libc resolves inside the AppDir.

The wrapper mechanism was tested end-to-end on a real Linux host before this ran in CI; `/proc/self/maps` confirmed the bundled libc is the one actually mapped, `getAppDir()` was unchanged, and the exported `HERE` that `bin/ruby` reads was passed through untouched.

`appimage-build` is green with this change: it bundles `glibc-2.40-224`, wraps all 11 dynamic executables and correctly skips the static `wazero`. **Artefact size: 341,402,056 -> 343,187,912 bytes (+1.7 MB, +0.5%).**

## Result: all five legs green (run 33809203244)

| distro | glibc | before | after |
|---|---|---|---|
| `ubuntu:22.04` LTS | 2.35 | fail | **pass** |
| `debian:12` stable | 2.36 | fail | **pass** |
| `fedora:40` | 2.39 | fail | **pass** |
| `ubuntu:24.04` LTS | 2.39 | fail | **pass** |
| `archlinux:latest` | ~2.42 | pass | **pass** (no regression) |

`appimage-build` and `appimage-lib-check` are green too — `appimage-lib-check` was failing before this branch.

**Proof the bundled libc is the one resolved.** On `ubuntu:22.04` (glibc 2.35) and `debian:12` (glibc 2.36) — hosts that cannot supply `GLIBC_2.38` or `GLIBC_ABI_GNU2_TLS` at all — the AppImage printed `CodeTracer version: 26.08.1`, and the probe reported

```
ld.so --list  : libc.so.6 => <AppDir>/lib/libc.so.6
```

**Artefact size: 341,402,056 -> 346,227,144 bytes (+4.8 MB, +1.4%).**

### Three unrelated CI defects had to be fixed to get any verdict at all

Both are on `dev`, neither is caused by this change, and between them they meant the AppImage could not be built *or* tested by anyone:

1. **`lint-nim` was red on `dev`** because `ci/test/constraints-listing-browser.sh` (from 2bf98864) was unrecorded in `shell-gate-coverage.known-dark.txt`. `appimage-build` and `dmg-build` both `need:` `lint-nim`, so the AppImage could not be built at all. Fixed in ba4447b3.
2. **Every cross-distro leg died at `aws: command not found`** in the Download step, before the AppImage was fetched — so the matrix reported five failures while testing nothing. Fixed here (`nix shell`, matching `appimage-lib-check`).
3. **This PR's own probe was measuring the wrong process.** `LD_DEBUG` on `bin/ct_unwrapped` attaches to the wrapper's *bash*, whose libc is the host's, so it failed two legs on an AppImage that was running correctly. Fixed to invoke the loader the way the wrapper does.

### DIFF
--- a/scripts/test-appimage-cross-distro.sh
+++ b/scripts/test-appimage-cross-distro.sh
@@ -218,9 +218,19 @@ case "$resolved" in
 	;;
 esac
 
-# (b) Reality: the same claim about the process that actually runs. LD_DEBUG
+# (b) Reality: the same claim about a process that actually runs. LD_DEBUG
 #     reports what the loader really mapped, not what it would have chosen.
-LD_DEBUG=libs "$APPDIR/bin/ct_unwrapped" --version >/dev/null 2>/tmp/lddbg.txt || true
+#
+#     Invoke the LOADER the way the wrapper does, rather than running
+#     $APPDIR/bin/ct_unwrapped. That path is now a bash wrapper, so LD_DEBUG
+#     would attach to bash first and the first "calling init:" line reported
+#     would be BASH's libc -- the host's -- which says nothing about the
+#     program. This check failed on debian:12 and ubuntu:22.04 for exactly that
+#     reason, on an AppImage that was in fact running correctly: `--version`
+#     printed the version on both, which glibc 2.35/2.36 could not have done
+#     had the host libc been the one in use.
+LD_DEBUG=libs "$LOADER" --library-path "$APPDIR/lib" "$TARGET" --version \
+	>/dev/null 2>/tmp/lddbg.txt || true
 initpath=$(grep -oE "calling init: [^ ]*libc\.so\.6" /tmp/lddbg.txt | head -1 | sed 's/calling init: //')
 echo "LD_DEBUG=libs : libc.so.6 => ${initpath:-<not reported>}"
 case "$initpath" in


zah:
Landed directly on `dev` as `cf503bc2` (rebased clean onto `b6e28026`, zero conflicts).

Verified green on run [33809203244](https://github.com/metacraft-labs/codetracer/actions/runs/33809203244) before landing: `appimage-build`, `appimage-lib-check`, and all five `appimage-cross-distro` legs — ubuntu:22.04, debian:12, fedora:40, ubuntu:24.04, archlinux:latest.

Closing; the commits are on `dev`.

PR #716, opened 2026-09-07
security(deps): 194 → 24 Dependabot alerts on stable, all 3 criticals cleared

zah:
## Measured, with the instrument calibrated first

`gh api` reports **194 open alerts on `stable`** — 3 critical, 92 high, 73 medium, 26 low. A verifier that re-derives alert state by matching lockfile versions against advisory ranges reproduces **194 remaining / 0 fixed / 0 unknown** on pristine `origin/stable`, so the scoring below is against a calibrated instrument rather than an assumption.

**Two premises that were widely believed and are false:**

- An earlier audit "corrected" the count to 183. It is **194**.
- "The fix already landed on `dev`" — `dev` carries **192 of the same 194**; only two are fixed there, and all three criticals are on `dev` too. Four of the six vulnerable manifests are byte-identical blobs between the branches. Promotion would close 2 of 194 while dragging in 978 commits; it is not a security action.

`stable` is the repo's default branch, which is why alerts are computed there — and why only a merge can move the number.

## Result: 194 → 24

165 fixed, plus 5 whose manifest was deleted. Remaining: **0 critical**, 11 high, 7 medium, 6 low.

The load-bearing find: twelve `node-tar` advisories are **unbounded below**, so `tar@6.2.1` matched every one including the critical, and bumping the 7.x pin changed nothing. The deprecated `electron-rebuild@3.2.9` was the *sole* path to it (`node-gyp@9 → cacache@16 → tar@6`); nothing in the repo invokes it, and its only other reference is an Electron-12-era line in `docs/walkthrough.md`. `@electron/rebuild@^4.2.0` pulls `tar@^7.5.4`, so `tar@6` leaves entirely.

Also: one recursive lockfile refresh with no `package.json` range changes (~98 npm alerts), `openssl 0.10.72 → 0.10.81`, and `test-programs/rs_stylus_vesting/` deleted — it held exactly one tracked file, a `Cargo.lock` with no `Cargo.toml` and no sources, orphaned by `af16ac382`. Cargo cannot consume it. A dead file, not a suppression.

`dompurify → 3.4.15` is **its own commit so it can be dropped independently** — it is the one change putting an untested sanitizer version under monaco, which pins it exactly.

No alert was dismissed and no `dependabot.yml` exclusion was added.

## The 24 that remain

**Unfixable — no patched version exists (3):** `extract-zip ≤2.0.1` and `ruint ≤1.17.0` both report `first_patched_version: null`; both unmaintained.

**Needs a major migration (21):** the Storybook 7 tree (11 — Storybook 7 is EOL, this is a 7→9 migration), `stylus_fund_tracker` (8 — a Stylus test fixture that ships nothing), `ip-address` (2, blocked by `socks@^2.8.3`), and `immutable` (1 — GHSA-v56q-mh7h-f735 publishes no 3.x backport).

## Expect one red round-trip, by design

`node-packages/yarn-project.nix` pins `cacheOutputHashes`, a fixed-output hash of the entire yarn cache per platform. Changing `yarn.lock` invalidates both entries, so `node-modules-derivation` **will fail until they are refreshed** — and they cannot be computed locally (no Linux builder configured; the `x86_64-linux` hash is the one CI and the binary cache build against).

Every yarn command was run with `YARN_ENABLE_NIXIFY=false` so the plugin would not clobber that hand-maintained attrset; `yarn-project.nix` is unmodified in this diff. The documented procedure applies: take the `got:` hash from each mismatch error, for `x86_64-linux` and `aarch64-darwin`.


zah:
Superseded by #737, which landed the tar critical on `dev`.

Measured on the tree rather than from the alert text:

- `electron-rebuild@3.2.9` was the **sole** root of the tar 6.x subtree. Replacing it removes **67 (package,version) pairs, adds 6, with zero downgrades**, leaving only `tar@7.5.22` — which clears all **12** tar alerts, not just the critical.
- This PR's Cargo.lock bumps are **already satisfied on `dev`**: the `bytes` alert range is `>= 1.2.1, < 1.11.1` and `dev` is already at 1.11.1. The two test-program lockfiles carry only *medium* `bytes` alerts.
- This PR's `storybook/package-lock.json` **still contains `tar@6.2.1`** — it did not fix tar there.
- It rewrites `yarn.lock` without touching `yarn-project.nix`, which carries the per-system hash block. Both hashes would be stale, which reports no mismatch and dies with `YN0056` — breaking every Nix build.

Since `stable` has now been merged into `dev`, promotion is the route for anything still wanted from here.

PR #718, opened 2026-09-07
feat(windows): make a toolchain tree survive being moved, and prove it by moving one

zah:
A toolchain tree is built on one machine and unpacked on another. A tree that records where it was built does not break when it moves — it keeps working, against directories that belong to the machine that made it, and every consumer after that inherits the same wrong answer without a single error message.

## The conversions

GCC published itself as a junction **unconditionally**, on every arm including the download-and-verify path that had already extracted the tree inside the install root. LLVM and zstd junctioned at a system install when they found one.

- **GCC** — the junction is gone on both arms. The download arm points at the tree it already extracted (`gcc/winlibs-<version>/mingw64`), which costs nothing. The winget arm **copies** into the install root rather than junctioning at a per-user winget directory: not free, and the right trade, since a junction there yields a tree that works only on the machine and account that installed it.
- **LLVM / zstd** — the system and `C:\zstd` arms became copies, keeping the saved download while producing a self-contained tree.
- All three record `install_arm` in `<component>.install.meta` and write `<component>.install.relative-path`. `env.ps1` reads all three back through `Resolve-InstallDirFromRelativePathFile`, with the historical layout as a fallback so an install root provisioned before this change still resolves.

## The audit judges what a link STORES, not where it resolves

These are different tests and the difference is the whole point. An NTFS junction records an **absolute** path, so `<root>\gcc\15.2.0 → <root>\gcc\winlibs-15.2.0\mingw64` — a target inside the install root — still names the *original* root once the tree is unpacked elsewhere. A resolution-based check clears exactly the shape that fails.

`Get-ReparsePointFindings` therefore returns three verdicts: `reparse-absolute-target` (a violation unconditionally), `reparse-escapes-root` (a violation), and `reparse-inside-root` (a relative target within the root — it travels, not a violation).

The reparse scan is paired with a **content** scan, because a component can be non-relocatable with zero reparse points: rustup records absolute paths as plain text. `Get-BakedAbsolutePathFindings` searches bytes rather than lines, in UTF-8 *and* UTF-16LE, across three spellings of the root, does not follow reparse points, and reports oversized or unreadable files as UNSCANNED rather than passed.

## The gate is a round trip, not a string comparison

`ci/test/windows-store-relocation.ps1`, **103 checks**, wired into `windows-bootstrap-smoke` (already a required job).

Five component trees are laid out at root R1 through the **production** `Write-InstallPointer`, proven to RUN there, archived with `tar`, unpacked at R2, and **R1 is then DESTROYED**. Each tool is resolved through its pointer and executed, prints its own resolved directory, and is asserted to be under R2 and **not** under the destroyed R1 — the assertion a string comparison on a pointer file cannot make. Per component, never in aggregate.

Its negative control bakes an absolute path: proven to work at R1, found by the content scan, **missed** by the reparse scan (so the two checks are demonstrably not redundant), and failing after relocation — so the round trip is known to be capable of failing.

**Eleven mutations of the production code each turn the suite red**, including the one that reverts to resolution-based judging.

## WARN → FAIL, and where it will trip first

`Assert-BootstrapRelocatability` throws rather than warns: under publish-and-refill a mis-declared component is not a note in a log, it is an entry every later job inherits. It is called **outside** the report's `finally`, so a relocatability throw cannot replace the error that actually stopped a provision.

It will trip first on the rustup tree, which still declares itself relocatable and whose reparse points nobody has characterised — the previous check was a bare count. The comment at the promotion site says so, and says the two correct responses are to convert that layout or change its declaration, **not** to lower the gate back to a warning.

## Also

- `Ensure-CleanDirectory` refuses any path at or under a content-addressed store root, naming both the path and the root. Both arms asserted: private scratch is still cleaned *and* verified to have really been emptied; `<root>-scratch` is not treated as a child of `<root>`; with no store configured nothing is fenced.
- `Repair-RustupSettingsRelocatability` drops `settings.toml`'s `[overrides]` table after install. `version`, `default_toolchain`, `profile` and any table *after* `[overrides]` are asserted to survive, so the repair is targeted rather than a truncation.
- `non-nix-build/windows/store-exclusions.json` records the FreePascal and MSVC decisions. The gate cross-checks it against **`env.ps1`'s own parsed dispatch block** rather than a second hand-maintained list — two lists can drift together. MSVC's exclusion is structural: absent from the dispatch entirely, and `export-msvc-env.ps1` asserted to contain no download, extract or install step (each absence checked, not the comment quoted). Both open FreePascal questions are recorded as **UNMEASURED**, each with why and with the exact procedure that would answer it; the gate asserts each carries a status and a method so the pair cannot decay into folklore.
- `ConvertTo-InstallRelativePath` had a platform-separator bug that rejected every path under the root on any non-Windows host, which is why this discipline had never been exercisable by the suites that run on Linux `pwsh`.
- `ensure-gcc.ps1`, `ensure-llvm.ps1` and `ensure-zstd.ps1` were missing from the PowerShell parser list and are added.

## Verification, and one limit stated plainly

`windows-store-relocation.ps1` 103/103 and `bootstrap-decomposition.ps1` 69/69 on Linux `pwsh` 7.5.4; the parser gate 19 files, 0 failures.

**The converted `Ensure-Gcc` / `Ensure-Llvm` / `Ensure-Zstd` have never executed.** The conversions are verified by the tree-walking audit and the round trip over trees shaped like theirs, not by a provisioning run. Closing that needs one green Windows provision.

`validate-install-root.ps1` fails on Linux for a pre-existing reason it declares itself: it needs real Windows NTFS ACLs. It is untouched by this change.

The first four commits are the Windows provisioning-probe work this builds on.


zah:
Superseded by a branch based on `dev`. This branch carries four earlier Windows provisioning-probe commits, and one of them (`b67f1de0f`) adds three branch-tip `siblings:` entries in `windows-provision-probe.yml`, which puts `ci-contract-suites` over its ceiling (59 vs 56). That is unrelated to the store-relocatability work and should not ride along with it. Reopening the same change against `dev` alone.

PR #727, opened 2026-09-08
ci: run every lane of `just test` and `test-bpf`, and name every failure

zah:
## The defect

Two aggregate recipes in `justfile` reported only their **first** failing lane.

**`just test`** (justfile:771) — this is the `test-non-gui` CI job, reached as
`codetracer.yml` → `ci/test/non-gui.sh` → `just test`. It was a bash body under
`set -e` running seven `just <lane>` calls in sequence, so the first non-zero
lane aborted the body and the other six never ran. The job went red naming one
lane; whoever picked it up fixed that one, pushed, and discovered the second.
One full round trip of CI — nix dev-shell startup included — per broken lane,
and no point at which anybody could see how much was actually broken.

**`test-bpf`** (justfile:1311) — the same defect in the other spelling, a
dependency list of four lanes. `just` aborts the whole invocation at the first
dependency that exits non-zero. That one is *not* a `set -e` artefact and no
shell flag fixes it; the aggregate had to stop being a dependency list.

## The fix

Both delegate to the new `ci/lib/run-just-lanes.sh`, which runs every lane in
the original order and then fails naming all of them with their own exit
statuses.

**Nothing is weakened.** Every lane still runs the same recipe, every exit
status is still load-bearing, and both aggregates still exit non-zero if any
lane failed. There is no `continue-on-error`, no allow-list, no way to mark a
lane advisory. An empty lane list is a hard error (exit 2), not a vacuous pass.

The `CODETRACER_RR_BACKEND_PATH` gate on `cross-test` is preserved **as a gate**
— `ci/test/non-gui.sh` sets the variable empty on purpose — but the lane is now
appended to the aggregated list, so when it does run its failure fails
`just test` like any other.

## Proof, on real lanes

Four real lanes cheap enough to run outside the dev shell, with a failure
injected into **position 3 of 4** (`test-python-version-alignment`, `exit 9`).
Note that `test-build-alignment` at position 4 fails *genuinely* here, for want
of the dev shell:

| | lanes that ran | exit | what the log names |
|---|---|---|---|
| old `set -e` shape | 3 of 4 | 9 | injected lane only — **lane 4's real failure never reported at all** |
| new runner | 4 of 4 | 1 | `test-python-version-alignment (exit 9)` **and** `test-build-alignment (exit 1)` |

The old shape hid a genuine failure behind an injected one. That is the whole
bug, reproduced.

## The contract suite

`ci/test/run-just-lanes-test.sh`, registered as a `lint_step` in
`ci/lint/bash.sh` (bash + `just`; no nix, no build, no network, seconds). It
drives the real `just` binary over throwaway recipes in a temp dir and asserts
the later lanes run, the aggregate still fails, and every failure is named —
28 assertions.

Two of its six sections reproduce the **original** shapes and assert they hide
the later lanes. If those ever start passing trivially, the suite has stopped
being able to see the difference it exists to measure, and the rest of it would
be green for the wrong reason.

Mutation-tested, both caught:

- restore fail-fast (`break` after the first failure) → 6 assertions fail
- run every lane but drop the `exit 1` (weakened gate) → 2 assertions fail

## One behavioural difference, asserted rather than glossed

Under `set -e` the body died with the failing lane's **own** status (e.g. 3);
the runner reports a uniform `1` for "some lane failed". Safe because the only
consumer `exec`s the recipe and CI tests it for non-zero — no caller anywhere
reads a specific exit code off these aggregates. Asserted in the suite so the
difference is recorded rather than discovered later.

## Survey: other aggregates of the same shape

Denominator: **208 recipes** parsed from `justfile` by a column-0 recipe-header
scan (independently corroborated — `ci/test/shell-gate-coverage.sh`'s own header
cites the same 208). Two shapes searched: headers with ≥2 dependencies, and
bodies invoking `just <lane>` ≥2 times.

Raw: 8 dependency-list + 6 body-sequence. Most are **legitimate fail-fast
pipelines**, not defects — `test-gui: build-once build-siblings`, the three
`ensure-ct-mcr ensure-ct-native-replay` prerequisite pairs, `storybook`,
`capture-docs-visual-page` — where a later step genuinely cannot run if an
earlier one failed. (`test-gui-prebuilt` is a false positive: its two `test-e2e`
calls are mutually exclusive `case` branches.)

**Three genuine remaining instances**, deliberately not fixed here:

| recipe | line | CI-reachable? |
|---|---|---|
| `test-vm: test-vm-native test-vm-js` | 2801 | **yes** — `codetracer.yml:3301`, and it is red in the current `dev` baseline |
| `test-flow-all` (17 language lanes, `set -e`) | 2603 | no |
| `test-nim-flow-all` (3 lanes, `set -e`) | 1686 | no |

`test-vm` is the highest-value one and would be a two-line conversion — **but it
sits at justfile:2801, inside the region PR #726 is appending ~87 lines to.**
Fixing it here would collide. Deferred on purpose; it should follow once #726
lands.

## Risk

`test-non-gui` is **already failing on both platforms** in the `dev` baseline at
`db2c88167`, so this cannot turn a green job red. Expected effect: that job's
log now names *multiple* failing lanes where it previously named one.

---

## Update: the `lint-bash` regression, and how the proof was obtained

The first version of this PR turned `lint-bash` from green to red. Fixed in `9e813e93a` — full detail in the PR comment below. Summary: the contract suite drove the real `just`, and `lint-bash` runs in `devShells.x86_64-linux.lint`, which omits `just` deliberately. The suite is now hermetic (stubs `just` on PATH); the fail-closed behaviour is unchanged and mutation-tested under the no-`just` condition.

### Caveat on how the hermetic proof was obtained — read this before trusting the green

I could **not** run the real lint shell locally: it is `x86_64-linux` and this workstation is darwin. So the hermetic runs below **simulate** `lint-bash` by emptying `just` from `PATH`, rather than by entering `devShells.x86_64-linux.lint`.

| scenario | assertions | exit |
|---|---|---|
| full (real `just`) | 28 | 0 |
| hermetic (simulated lint shell: no `just` on PATH) | 25 | 0 |
| hermetic + fail-fast restored | **6 FAIL** | 1 |
| hermetic + gate removed | **2 FAIL** | 1 |

That is a proxy, not the real shell. The residual risk it leaves: if `lint-bash` still fails, the likeliest cause is a tool the suite's stub path needs that the lint shell lacks. It uses only `mktemp`, `chmod`, `cat` and `rm` — all from `coreutils`, which `nix/shells/lint.nix` declares — so the risk is believed low, but it is **not** discharged by the table above. The authority is the CI run, not this table.

### One section is knowingly DARK IN CI

The section asserting what `just` **itself** does with a failing dependency needs the real binary, so it does not run in `lint-bash`. It now says so **in the log**, naming the missing capability and what would lift it.

It is **not** registered in `ci/test/shell-gate-coverage.known-dark.txt`, deliberately: that list is keyed on whole **gates no lane can reach** and fails in both directions — an entry that becomes reachable fails the run by name. This script *is* reachable and stays reachable, so a line for it would be false the day it was written; its ceiling is also reserved for BLOCKED gates. `ci/lib/known-test-failures.tsv` does not fit either — it registers cases that FAIL with a signature, and requires the failing set to equal the registered set exactly; this section does not fail, it does not run.

**No registry in the tree has section granularity.** That is reported as a gap rather than resolved by inventing a fourth registry.

### DIFF
--- a/ci/lib/run-just-lanes.sh
+++ b/ci/lib/run-just-lanes.sh
@@ -1,0 +1,127 @@
+#!/usr/bin/env bash
+# shellcheck shell=bash
+#
+# run-just-lanes.sh — run EVERY lane of a `just` aggregate, then fail naming
+# ALL the lanes that failed, not just the first one.
+#
+# WHY THIS EXISTS
+# ---------------
+# An aggregate recipe is a promise: "`just test` runs the non-GUI tests". Two
+# ways of writing one in a justfile both quietly break that promise, and both
+# were in this file's tree when it was added:
+#
+#   * A BASH BODY UNDER `set -e`:
+#
+#         test:
+#           #!/usr/bin/env bash
+#           set -e
+#           just test-build-alignment
+#           just test-flake-pin-alignment
+#           ... five more ...
+#
+#     `set -e` aborts the body at the FIRST non-zero lane. The other six never
+#     run. The CI job (`test-non-gui`, via ci/test/non-gui.sh) goes red naming
+#     one lane, so whoever picks it up fixes that one, pushes, and discovers the
+#     second — one round trip of CI per broken lane, each one costing the full
+#     nix dev-shell startup, and no point at which anybody can see how much is
+#     actually broken.
+#
+#   * A DEPENDENCY LIST:
+#
+#         test-bpf: test-bpf-monitor test-bpf-native test-bpf-native-integration test-bpf-integration
+#
+#     Identical failure mode for a different reason: `just` runs dependencies in
+#     order and aborts the whole invocation when one exits non-zero. It is not a
+#     `set -e` artefact and cannot be fixed by changing shell flags; the only fix
+#     is to stop expressing the aggregate as a dependency list.
+#
+# Neither shape is wrong about WHETHER to fail. Both are wrong about WHAT THEY
+# REPORT: an aggregate whose job is to answer "is the non-GUI suite healthy?"
+# answers "the first thing I tried is not", which is a strictly smaller and less
+# useful fact. This script answers the actual question.
+#
+# WHAT IT DOES NOT DO
+# -------------------
+# It does not weaken anything. Every lane still runs the same recipe, its exit
+# status is still load-bearing, and the aggregate still exits non-zero if ANY
+# lane failed. There is no continue-on-error, no allow-list, no way to mark a
+# lane advisory. The only change is that a failing lane no longer suppresses the
+# lanes after it.
+#
+# It also refuses a VACUOUS PASS. Zero lanes is not "everything passed"; it is a
+# caller that built an empty list, and the whole point of this area of the tree
+# (see ci/lib/run-nim-test-lane.sh, ci/test/shell-gate-coverage.sh) is that a
+# gate which cannot fail is indistinguishable from one that is not running.
+#
+# ORDER IS PRESERVED. Lanes run sequentially in the order given, exactly as they
+# did before, because some of them share build outputs and were written assuming
+# that. This is not a parallel runner.
+#
+# Usage:
+#   bash ci/lib/run-just-lanes.sh <aggregate-name> <lane> [<lane> ...]
+#
+# Exit status:
+#   0  every lane passed
+#   1  at least one lane failed (all failures named on stderr)
+#   2  usage error, or no lanes given
+#
+# Environment:
+#   CT_JUST  the `just` binary to invoke (default: `just`). Exists so
+#            ci/test/run-just-lanes-test.sh can point the runner at a
+#            throwaway justfile; not used in CI.
+
+set -uo pipefail
+
+aggregate="${1:-}"
+if [ -z "${aggregate}" ]; then
+	echo "usage: run-just-lanes.sh <aggregate-name> <lane> [<lane> ...]" >&2
+	exit 2
+fi
+shift
+
+just_bin="${CT_JUST:-just}"
+
+if [ "$#" -eq 0 ]; then
+	echo "run-just-lanes.sh: '${aggregate}' was given NO lanes to run." >&2
+	echo "  Refusing to report success for an empty aggregate: a gate that" >&2
+	echo "  cannot fail is not a gate. Fix the caller's lane list." >&2
+	exit 2
+fi
+
+lanes=("$@")
+passed=()
+failed=()
+
+for lane in "${lanes[@]}"; do
+	echo ""
+	echo "=== ${aggregate}: lane ${lane} ==="
+	status=0
+	"${just_bin}" "${lane}" || status=$?
+	if [ "${status}" -eq 0 ]; then
+		passed+=("${lane}")
+		echo "=== ${aggregate}: lane ${lane} PASSED ==="
+	else
+		failed+=("${lane} (exit ${status})")
+		# Announced on stderr as well as stdout so that a reader tailing only
+		# the error stream still sees every failure, not merely the summary.
+		echo "=== ${aggregate}: lane ${lane} FAILED (exit ${status}) ===" >&2
+		echo "    continuing — remaining lanes still run, and this failure is" >&2
+		echo "    recorded and re-reported in the summary below." >&2
+	fi
+done
+
+echo ""
+echo "=== ${aggregate}: summary — ${#lanes[@]} lane(s): ${#passed[@]} passed, ${#failed[@]} failed ==="
+
+if [ "${#failed[@]}" -gt 0 ]; then
+	echo "" >&2
+	echo "${aggregate}: ${#failed[@]} of ${#lanes[@]} lane(s) FAILED:" >&2
+	for entry in "${failed[@]}"; do
+		echo "  - ${entry}" >&2
+	done
+	echo "" >&2
+	echo "  Every lane above was run; this list is complete, not first-failure." >&2
+	exit 1
+fi
+
+echo "${aggregate}: all ${#lanes[@]} lane(s) passed"
--- a/ci/lint/bash.sh
+++ b/ci/lint/bash.sh
@@ -169,10 +169,19 @@ lint_step "shellcheck: build-alignment harness ('just test' runs it)" \
 #
 # Executed here, and not only linted, for the reason the stale-artefact suite
 # below is: an aggregate that has never been SEEN to report a second failure is
-# indistinguishable from one that still hides it. The suite drives real `just`
-# over throwaway recipes in a temp dir -- no nix, no build, no network, seconds
-# -- and it reproduces BOTH original shapes to prove it can still tell the
-# difference.
+# indistinguishable from one that still hides it.
+#
+# It is HERMETIC — it stubs `just` on PATH rather than needing the real one —
+# and that is a correctness requirement of running it here, not a convenience.
+# nix/shells/lint.nix carries no `just` on purpose, so the first version of this
+# registration refused to run and turned this job red. The stub costs nothing:
+# ci/lib/run-just-lanes.sh's whole interface to the outside is `just <lane>`,
+# one argument and one exit status. So: no nix, no build, no network, seconds.
+#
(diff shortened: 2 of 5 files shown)


zah:
## Fixed the `lint-bash` regression (9e813e93a)

`lint-bash` was **green on `dev`** (run 34247470211 @ db2c88167) and **failed on this PR** (run 34279362460, job 102240127154). That is a regression by set inclusion and it was mine. Fixed.

### Cause — and the diagnosis my own error message got wrong

The suite drove the real `just`. `lint-bash` runs `ci/lint/bash.sh` inside `devShells.x86_64-linux.lint`, and **`nix/shells/lint.nix` carries no `just`, deliberately** — its package list is derived from "every command in command position across the lint scripts", and its header says a future need is "a signal to put it in another shell rather than to widen this one".

To answer the question directly: **`just` IS declared in `nix/shells/ci-base.nix:185`** — my claim was true. But `lint-bash` does not use ci-base. So my message ("its absence means this suite is running outside the shell it is specified for") sent the reader hunting for a broken shell when the real fault was a **misplaced suite**. The message is gone.

### Fix: option (a), hermetic

`ci/lib/run-just-lanes.sh`'s entire interface to the outside is `just <lane>` — one argument, one exit status. A stub first on `PATH` exercises everything that is ours: ordering, per-lane exit-status capture, failure accumulation, verdict. What a stub cannot exercise is whether `just` itself dispatches a recipe correctly — not this repo's code, and not what any assertion here claims.

The suite now **asserts the stub is the `just` it resolved**, so a mis-resolution cannot leave the marker-based assertions measuring something else.

Adding `just` to the lint shell would have been wrong twice over: it widens the shell that exists to be narrow, to serve a test that does not need the tool.

**The fail-closed behaviour is not weakened.** Nothing is skipped, quarantined, or made advisory.

### The one honest limitation

Section 6 asserts what **`just` itself** does with a failing dependency — the fact that forced `test-bpf` to stop being a dependency list. A stub cannot serve that. It self-skips without a real `just`, and the expected assertion count drops 28 → 25 so **a short tally is still a finding in both modes**. Stated plainly in the file: **in CI today that section is exercised by no lane.** The gate proper — the sections asserting what the runner does now — is fully hermetic and always runs.

### Measured, including both mutations re-run under the hermetic mode `lint-bash` will use

`PATH` with no `just`, absence verified separately:

| scenario | assertions | exit |
|---|---|---|
| full (real `just`) | 28 | 0 |
| **hermetic (no `just`)** | **25** | **0** |
| hermetic + fail-fast restored (`break` after first failure) | **6 FAIL** | 1 |
| hermetic + gate removed (runs every lane, never fails) | **2 FAIL** | 1 |

The stub does not cost the suite its power to fail closed.

`shellcheck` 0.11.0 clean, `shfmt -l -ln auto -s` clean, `just --summary` parses.

---

## The second failure is NOT mine

`contract suite: existence is not freshness` → `a rebuilt renderer makes the copied corpus stale / output: FRESH`. Investigated rather than assumed. It is a **pre-existing latent mtime race** in `ci/test/stale-artefact-guards-test.sh`:

- The guard compares **strictly**: `stat.mtimeMs > referenceMs` (`tools/visual-review/storybook-freshness.mjs:42`), reference `statSync(built).mtimeMs` (line 77). **Equal mtimes ⇒ `FRESH`** — exactly the observed output.
- Lines 403 and 408 are **two back-to-back `touch`es with zero intervening work** — no `sleep`, no `touch -t`. Typically 1–3 ms apart, inside one kernel coarse-clock tick (≈1–10 ms, whole seconds on some filesystems).
- **The suite already knows this hazard and guards against it elsewhere in the same file** — lines 1151–1156 use explicit `touch -t` mtimes with the comment *"two files written in the same second would make this pass for the wrong reason"*. The storybook section is the outlier.
- `storybook-freshness.mjs` is **four days old** (single commit `663c8ab16`); no commit in the suite's history mentions flake/mtime/granularity. New assertion, never characterised, first observed red now — the profile of a latent race, not a regression.

No data, control, cwd, tempdir or state path connects it to my four files, and `ci/lib/lint-steps.sh:163` runs every step in a **subshell** `("$@")`, so an inserted step cannot leak into it. The only coupling is wall-clock/scheduling — a *reveal*, not a *cause*.

**Not fixing it here** (it is outside this change and I was asked not to touch renderer code). Recommended fix for whoever owns it: give lines 403/408 explicit ordered mtimes the way lines 1151–1156 already do.

zah:
## Final state — `lint-bash` green, aggregate fix proven on real CI

### `lint-bash`: regression cleared

Job `102261628118`, `conclusion=success`, verified by direct API read rather than a reported status. Internal summary `44 step(s): 44 OK, 0 FAILED, 0 QUARANTINED`. Both previously-failing steps are green, and the hermetic suite reported:

```
run-just-lanes-test: mode=hermetic
run-just-lanes-test: OK — 25 assertions, 0 failures
```

with section 5 still reproducing the original defect (`old shape: lane-c did NOT run — the defect, reproduced`). The instrument is demonstrably alive, not vacuously passing.

### The aggregate fix, proven on real infrastructure

`test-non-gui (macos, aarch64-darwin)`, identity confirmed via `Complete job name` in both logs:

| | baseline (`dev` @ `db2c88167`, job `102210870173`) | this PR (job `102261628575`) |
|---|---|---|
| occurrences of `nimsuggest` in the whole log | **0** | **29** |
| outcome | `Recipe test-rust failed with exit code 100` → `Recipe test failed` | `test: 1 of 7 lane(s) FAILED: - test-rust (exit 100)` |

Same platform, same lane, same exit 100, same root cause (`tracepoint_interpreter` tests). `test-rust FAILED` at log line 2865; `test-nimsuggest` starts at 2869 and passes. **The lane after the failure now runs and is reported, where on `dev` it never executed at all.**

Non-regression by set inclusion: `failure` → `failure`, same cause. Not a new red.

### The freshness race — read this before trusting the green

**That green `lint-bash` run passed `existence is not freshness` WITHOUT the race fix. It won the race that time.** That is *not* evidence the race is gone, and it is precisely why a rerun can never settle this. The unmodified suite also passed 3/3 locally before any change was made.

The only thing that separates the hypotheses is the deterministic repro: give the corpus and `ui.js` the **same** stamp on purpose, which the strict compare (`stat.mtimeMs > referenceMs`) must read as `FRESH`.

- **Before the fix:** 3 of 3 runs FAILED, each with the exact CI signature (`the command unexpectedly succeeded / output: FRESH`).
- **After the fix:** 3 of 3 runs PASSED, `expected=107 executed=107 failed=0` — assertion count unchanged.
- **Teeth check:** mutate the fix so the renderer stamp *equals* the rebuild stamp → the assertion FAILS again. A fix that merely silenced it would have stayed green.

Stated as strong inference, not proof: the CI log shows only `FRESH` and never records the two mtimes.

### Section 6 is untested in every CI job — plainly

Verified by enumeration, not assumption. Exactly one invocation of the suite exists in the tree (`ci/lint/bash.sh:186`); it runs in `lint-bash`, which uses `devShells.x86_64-linux.lint`; that shell declares `just` **0** times (`ci-base.nix` declares it once, but no lane runs this suite there).

**So the dependency-list instrument check — the assertion that `just` aborts at the first failing dependency, which is *why* `test-bpf` could not stay a dependency list — runs in NO CI job.** It is verified locally only (full mode, 28 assertions, exit 0). It self-skips with the expected count dropping 28 → 25, so a short tally is still a finding, and it now declares itself `DARK IN CI` in its own log with the missing capability named.

It is not registered in `ci/test/shell-gate-coverage.known-dark.txt`: that list is keyed on whole gates *no lane can reach* and fails in both directions, so an entry for this reachable script would be false the day it was written. No registry in the tree has section granularity — reported as a gap rather than papered over.

### Confirmations

- **Not merged.** PR #727 `state=OPEN`, `mergedAt=null`.
- **Nothing pushed to `dev`.** `origin/dev` is still `db2c88167`.
- Pushed once; remote head `b8fe4598b` verified via `git ls-remote`.

PR #729, opened 2026-09-08
ci: hoist viewmodel-tests' sibling clones above the App token's 60-minute expiry

zah:
## The defect

A GitHub App installation token expires **60 minutes** after it is minted.
`viewmodel-tests` mints exactly one, in its first step, and then performed three
authenticated clones — one at step 3, and two near the end of a job whose
`timeout-minutes` is 75. The last two were unauthenticated whenever the job took
longer than an hour to reach them.

Measured, in run `34219352454` / job `102038547352`:

| time | step | token age | result |
|---|---|---|---|
| 12:28:05 | `Generate CI token` | — | minted |
| 12:29:32 | clone `codetracer-trace-format-nim` | +1.5m | **success** |
| 13:32:28 | clone `codetracer-js-recorder` | +64.5m | `git exit 128` |
| 13:32:34 | clone `runquota` | +64.5m | `git exit 128` |

both with `fatal: could not read Username for 'https://github.com': terminal
prompts disabled`, taking the four steps that depend on those siblings down with
them. Nothing about the clones changed — the job grew past an hour and started
reaching an expiry that had always been there. It was invisible for as long as
this job died at ~20 minutes.

## The fix

Both late clones move up beside the step-3 clone. All three, plus the
`Install Nix` step that also spends the token, now run within the first couple of
minutes of the token's life. No `continue-on-error`, no skip, nothing weakened.

### Why not refresh the token instead

That was the first attempt, and it is what `metacraft-labs/reprobuild`'s `test`
job does — it mints four times (`Refresh installation token for Reprobuild
bootstrap` / `... for the CodeTracer build` / `... for tests`). **It is not
available in this repo.** `ci/test/sibling-provisioning-test.sh` asserts *"no job
mints more than one installation token"*, and a second `create-github-app-token`
step fails it:

```
  FAIL no job mints more than one installation token
     codetracer.yml:3397: job 'viewmodel-tests' mints a second token (already mints: ci_token)
```

Weakening that contract to fix a clone-ordering problem would trade a guard
against two interchangeable credentials for a convenience. The clones move
instead, and a comment block above them records the constraint so the next person
does not rediscover it from a red run.

### Safety of the move

Neither clone has an ordering dependency on anything between the old and new
positions: `clone-repo` runs plain git; `recorder_gate.nim` keys on the **built**
recorder binary rather than on the checkout existing, so no suite changes
behaviour by finding the sibling on disk earlier; and
`provision-repro-lock-siblings` only reconciles the five repos in `repro.lock`'s
`depends`, which include neither of these.

Verified locally under bash 5.3 (all exit 0): `sibling-provisioning-test.sh`
27/27, `recorder-clone-implies-build-test.sh` 14/14 — still passing both
*"viewmodel-tests is not reported: its Justfile route is followed"* and its
negative control *"a build that runs before the clone does not count"*, so the
clone is still ordered before the build — plus `job-timeouts-test.sh`,
`nix-provisioning-test.sh`, `origin-dap-gate.sh`,
`readonly-leftovers-sweep-test.sh`, `build-once-workspace-test.sh`,
`direnv-provenance-test.sh`, `grep-q-pipefail-gate.sh`.

## How widespread is this? (org-wide sweep)

Counts and method are in the review comment below.

### DIFF
--- a/.github/workflows/codetracer.yml
+++ b/.github/workflows/codetracer.yml
@@ -2730,6 +2730,75 @@ jobs:
           gh-token: ${{ steps.ci_token.outputs.token }}
           submodules: 'false'
 
+      # THE THREE SIBLING CLONES ABOVE AND BELOW ARE DELIBERATELY ADJACENT.
+      #
+      # A GitHub App installation token expires 60 MINUTES after it is minted.
+      # `Generate CI token` above is this job's ONLY mint -- `ci/test/
+      # sibling-provisioning-test.sh` asserts that "no job mints more than one
+      # installation token", so refreshing it mid-job is not available here --
+      # and this job's `timeout-minutes` is 75. Every authenticated git
+      # operation must therefore happen near the TOP of the job.
+      #
+      # `codetracer-js-recorder` and `runquota` used to be cloned next to the
+      # steps that consume them, ~64 minutes in. In run 34219352454 /
+      # job 102038547352 the token was minted at 12:28:05, the
+      # `codetracer-trace-format-nim` clone at 12:29:32 succeeded, and those
+      # two clones -- reached at 13:32:28 and 13:32:34 -- both died with
+      #
+      #   fatal: could not read Username for 'https://github.com':
+      #   terminal prompts disabled
+      #
+      # taking the four steps that depend on those siblings down with them.
+      # Nothing about the clones changed; the job simply grew past an hour and
+      # started reaching an expiry that had always been there.
+      #
+      # Neither clone has an ordering dependency on anything between here and
+      # where it used to sit: `clone-repo` runs plain git, the recorder gate in
+      # src/frontend/viewmodel/tests/unit/recorder_gate.nim keys on the BUILT
+      # recorder binary rather than on the checkout existing, and
+      # `provision-repro-lock-siblings` only reconciles the five repos named in
+      # repro.lock's `depends`, which include neither of these.
+      #
+      # Keep new clone steps in this block. A clone added further down is a
+      # clone that works only until the job gets slower.
+      - name: Clone codetracer-js-recorder sibling
+        # The recorder-gated ViewModel tests
+        # (src/frontend/viewmodel/tests/unit/test_{column,formatted_view_step,
+        # statement_step}_*_vm.nim) drive the real JS recorder to produce a
+        # column-aware trace and replay it through headless_session.  Without
+        # the sibling on disk they all hit recorder_gate's
+        # ``MISSING-RECORDER SKIP:`` path, so the column-aware / formatted-view
+        # / statement-step coverage silently vanished from CI.  Clone it into
+        # the workspace-sibling layout the tests resolve
+        # (../codetracer-js-recorder/packages/cli/dist/index.js) per
+        # codetracer-specs/Testing/Cross-Repo-CI-Integration.md.
+        uses: metacraft-labs/metacraft-github-actions/clone-repo@dev
+        with:
+          repo: metacraft-labs/codetracer-js-recorder
+          # `dev` is that repo's default branch; `main` trails it (15 commits
+          # behind when this was corrected), so the recorder-gated ViewModel
+          # suites and the `ct record` dispatch tests below were exercising a
+          # stale recorder while reporting green.
+          ref: dev
+          path: ${{ github.workspace }}/../codetracer-js-recorder
+          gh-token: ${{ steps.ci_token.outputs.token }}
+          submodules: 'false'
+
+      - name: Clone runquota sibling
+        # The `just build-once` step below compiles `ct`, whose import chain
+        # (src/ct/codetracer.nim -> ../ct_test/ct_test ->
+        # src/ct_test/process_exec.nim) starts with `import runquota_process`.
+        # Under tup the checkout is the only way those modules resolve --
+        # NIM_REPO_PATH_FLAGS points at `$(ROOT)/../runquota/...` and tup's
+        # environment allowlist omits RUNQUOTA_SRC (issue #641).
+        uses: metacraft-labs/metacraft-github-actions/clone-repo@dev
+        with:
+          repo: metacraft-labs/runquota
+          ref: dev
+          path: ${{ github.workspace }}/../runquota
+          gh-token: ${{ steps.ci_token.outputs.token }}
+          submodules: 'false'
+
       - name: Install Nix
         uses: ./.github/actions/setup-nix
         with:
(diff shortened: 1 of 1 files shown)


zah:
## How widespread is the expired-token defect? An org-wide measurement

The defect class: *a step that spends a GitHub App installation token on a remote,
executed more than 60 minutes after the mint that produced the token it spends.*
Measured from the **nearest preceding mint**, not from job start — measuring from
job start over-reports every job that already refreshes mid-run.

### Enumeration method (all EXECUTED, not reasoned)

1. `gh api orgs/metacraft-labs/repos --paginate` → **267** repos.
2. `gh api search/code -f q='org:metacraft-labs create-github-app-token path:.github/workflows'`
   → 114 files in **47** repos. *Caveat: GitHub code search indexes only the
   DEFAULT branch. codetracer's default is `stable`, where 6 files match; on `dev`
   the real number is 8. So 47 is a **lower bound** on repos and the file counts
   are stale for any repo whose default branch trails its mainline.*
3. For each of the 47, downloaded `.github/workflows/*` preferring the `dev`
   branch where it exists, else the default branch → **282** files.
4. Static pass over those files for jobs that mint a token **and** later spend it
   on a remote (token reaching a step via `with: gh-token/token/github-token`
   **or** via `env:` — see the blind-spot note below).
5. Measured pass: for every repo, the last 20 workflow runs (last **100** for
   `codetracer`, `reprobuild`, `io-mon`), pulling `/actions/runs/{id}/jobs` and
   keeping every job whose measured duration exceeded 60 minutes, with all step
   timestamps → **1062** unique long jobs across **42** repos.
6. Joined the two on exact step name within a repo, computing each
   token-spending step's age against its own nearest preceding mint.

### Counts, against stated denominators

| quantity | value |
|---|---|
| metacraft-labs repos (API listing) | 267 |
| …that mint an App token in a workflow (code search, default branch only) | **47** |
| workflow files parsed in those repos | 282 (7 unparseable — see below) |
| (file, job) pairs that mint **and** spend the token on a remote | **199** |
| …whose job ceiling exceeds 60 min (`timeout-minutes` > 60, or absent ⇒ 360 default) | **172** |
| unique jobs **observed** running > 60 min in the sampled runs | **1062** (42 repos) |
| …that actually executed a token-spending step | **32** distinct (repo, job) pairs |
| **jobs confirmed spending a token older than 60 minutes** | **1** — `viewmodel-tests`, this PR |

Token age at the moment a token-spending step started, across all 99 such
executions inside long jobs:

| age band | executions |
|---|---|
| 0–30 min | 94 |
| 30–45 min | 3 |
| 45–60 min | 0 |
| **> 60 min** | **2** (both in job `102038547352`, both fixed here) |

Closest approaches after this PR's job: `codetracer / windows-rust-components`
→ `Setup dev env` at **+34.9m**, and `codetracer / test-ui-tests (nixos)` →
`Clone codetracer-js-recorder sibling` at **+17.9m**. Nothing else in the corpus
gets past +6m. **The population is genuinely one.**

### The instrument was verified alive

A zero here would be meaningless on its own, so the same analyser was run against
a control corpus containing job `102038547352`. It returns exactly the two known
failures (`Clone codetracer-js-recorder sibling` +64.4m, `Clone runquota sibling`
+64.5m) and nothing else. The org-wide result uses the identical code path.

Two false positives were found and removed by tightening the instrument, which is
worth recording because both would have inflated the count:

* **`Post <step>` cleanup hooks.** `Post Checkout` legitimately runs at job end
  (up to +300m in `codetracer / queue-watchdog`) but only unsets a local
  credential — it contacts no remote.
* **Regex-on-step-name.** `reprobuild / Test` step `Build codetracer ct (via
  reprobuild)` matched a name regex at +74.8m, but that job **already refreshes
  its token four times** (`Refresh installation token for Reprobuild bootstrap` /
  `... for the CodeTracer build` / `... for tests`) and the step sits one line
  after a fresh mint. Exact-name joining against the static mint list removed it.

### One latent instance, unconfirmable by measurement

`codetracer / origin-dap-windows` (and `origin-dap-windows-nightly`) has the
identical shape: one mint at step 2, `timeout-minutes: 75`, and at step 10
`Checkout locked Python recorder adjacent to CodeTracer` — a `run:` step
receiving `SIBLING_TOKEN: ${{ steps.ci_token.outputs.token }}` and doing an
authenticated clone inside `ci/checkout-locked-python-recorder.ps1`. The job has
been measured concluding at **78.1 minutes** (job `102110643414`) with that
step's slot at **+74.3m**.

It cannot be *confirmed* because in all 6 sampled occurrences the job died at an
earlier step and this one is reported `skipped` — and a skipped step's
`started_at` is the job's end time, not a real start. That is exactly the shape
`viewmodel-tests` had until it survived long enough to reach its own late clones.

**It is deliberately not fixed in this PR.** The only order-preserving fix is a
second mint, which `ci/test/sibling-provisioning-test.sh` forbids; the
alternative is reordering steps 8/10 above the toolchain steps, which is a
larger change to a job on the `eph-win-x64` class that loses ~50% of its jobs to
zero-step provisioning failures and so cannot be verified by running it. Filing
it as a known latent defect is more honest than shipping an unverifiable reorder.

### Blind spots — stated, not hidden

* **Code search sees default branches only.** 47 repos is a lower bound. A repo
  that added a mint on a non-default mainline is invisible to step 2.
* **7 workflow files were not statically analysed** — all of them
  `ci-reprobuild.yml`, the same file replicated across `codetracer-{leo, move,
  polkavm, python, solana, ton}-recorder` and `codetracer-shell-recorders`.
  PyYAML rejects a PowerShell here-string whose embedded Python sits at column 0;
  GitHub's own parser accepts it (that workflow has 41 recorded runs), so this is
  an instrument limitation, not a repo defect. **Of the 31 sampled >60-min jobs
  in those 7 repos, 0 came from that workflow**, so the gap hid no measured long
  job — but it is unmeasured, not zero.
* **The static token-detector initially missed `env:`-passed tokens.** It was
  found only because `origin-dap-windows` turned up in the measured pass. Any
  other channel for handing a token to a step (a file, a composite action's
  internals) is still undetected.
* **Measured coverage is 46 of the 47 repos, and the 47th is unmeasurable.**
  `codetracer-example-recordings` returned 0 workflow runs — checked directly,
  `total_count: 0`, it has never run a workflow at all. Its 7 `regen-mcr-*` /
  `regen-rr-*` jobs carry no `timeout-minutes` (⇒ 360-min default) and clone
  siblings at steps 2/6/7, so they are counted in the **static** at-risk
  population above and are absent from the measured one by construction, not
  by oversight. A workflow that has never run cannot have exhibited the defect,
  but it also cannot be cleared by measurement.
* **Sampling depth.** 20 runs per repo (100 for the three named ones) from the
  API's recency-ordered listing — not full history. A job that crossed 60 minutes
  once, months ago, is not in this corpus.

### What is EXECUTED vs REASONED

**Executed:** the repo listing, the code search, the workflow downloads, the
static analysis, the 1062-job measured sweep, the join, the control validation,
and the eight contract tests run locally under bash 5.3.
**Reasoned:** that `origin-dap-windows` would 401 if it reached step 10 (its
step has never been observed running); that the 45–60 min band being empty means
no other job is near the cliff (true of the sample, not of all history).

PR #733, opened 2026-09-09
Stop reading the retired meta.json and paths.json from recordings

zah:
## What

Stops reading the retired container-internal `meta.json` / `paths.json`. A
recording's metadata and source paths come from `meta.dat`.

Completes the removal begun in metacraft-labs/codetracer-trace-format-nim#5
and metacraft-labs/codetracer-trace-format#58. **Order matters and this is the
back half:** those PRs stop every writer emitting the files; this one stops
the readers looking. Doing it the other way round breaks recordings in flight.

## Three call sites

1. **Follow-mode finalization** (`ctfs_trace_reader/ctfs_container.rs`).
   `FINALIZATION_META_FILES` listed both names; a follow reader treats the
   metadata file appearing as "the writer is done". Now `meta.dat` alone.

   This one was **only safe because of the writer change**: `meta.dat` used to
   be written only when a recording had at least one dedicated stream, so a
   splits-off recording sealed with `meta.json` and nothing else. Had this
   landed first, a follow reader would have polled such a recording forever.
   The writer PR made `meta.dat` unconditional; this depends on it.

2. **Source-path import** (`ct/trace/ctfs_sources.nim`). Dropped the read of
   the container-internal `paths.json`; paths come from `paths.dat` or
   `meta.dat`, both of which were already handled.

   **Careful here — two different files share the name.** The
   `outputFolder / "paths.json"` this writes is the *sidecar the frontend
   reads* to build the filesystem jstree, and it is untouched. Only
   `reader.readCtfsFile("paths.json")` — the container-internal one — is gone.
   A textual sweep hits the sidecar far more often than the target, which is
   why the surviving code now says so at the call site.

3. **Stylus flow test** — read the program name from `meta.json`; reads
   `meta.dat` via `parse_meta_dat` now. Only `program` was consumed, so the
   loader returns `MetaDat` rather than a serde `TraceMetadata`.

## Verification

- `cargo check --tests` on `src/db-backend` — clean.
- `nim check src/ct/trace/ctfs_sources.nim` — clean.
- `test_followfilesource_finalization_signal` re-pointed at `meta.dat`,
  including its negative arm (size 0 ⇒ not finalized).

I have **not** run the full db-backend test suite here; the two Rust changes
are a one-element const and a test-local loader, and both compile clean.

## Left alone deliberately

- `test_base40_roundtrip` still round-trips the string `"meta.json"`. It tests
  base40 encoding of a 9-character name, and old containers still carry such
  names — the encoding must keep decoding them.
- `online_sharing/test_mcr_enrichment.nim` writes a file called `meta.json`
  purely as an example of a **non-`.ct` file** the enrichment scanner must
  ignore. Unrelated to the container-internal sidecar.
- `trace.json` / `trace_metadata.json` / `trace_paths.json` — the
  pure-recorder cross-validation oracle shape — is a different legacy and is
  explicitly retained.


zah:
Superseded by #743.

This repository allows only rebase-merge. Landing this branch that way would replay `e42e485` and `ac0cec4` onto `dev` and drop the merge commit `b7b734b`, which is where the conflict resolution lives — including moving the correlation index off `meta.dat` bit 14, which `dev` has since given to the per-file line-count table. Both constants would then read `0x4000`, on different lines and with no textual conflict.

#743 carries the byte-identical tree as a single commit on top of `dev`, so a rebase-merge is a plain fast-forward. Closing this one once #743 lands.

zah:
Landed as #743 (merged as 8d526aa9). Same tree, restructured as a single commit so this repository's rebase-only merge could not drop the conflict resolution.

PR #737, opened 2026-09-09
security(deps): close the last critical (tar 6.2.1, GHSA-23hp-3jrh-7fpw) on dev

zah:
Closes the last open critical Dependabot advisory, on `dev` — where fixes belong. This is the `dev`-targeted replacement for #716, which targets `stable`.

## The advisory is not scoped to 7.x

`GHSA-23hp-3jrh-7fpw` (CVE-2026-59873, fix 7.5.19) has range **`<= 7.5.18`, with no lower bound**, so `tar@6.2.1` matched it. A claim in the tree that the tar advisories are scoped to 7.x — and that 6.x is therefore clean — is **false**, and the repo's own data disproves it without needing to trust a reading of the range:

`storybook/package-lock.json` contains **exactly one** `tar`, version **6.2.1, with no 7.x anywhere** (population: every key in `.packages` whose final path segment is `tar`; count 1, on both `dev` and `stable`). That manifest nonetheless carries **8 open tar alerts**, every one ranged `<= 7.5.x`. Ranges that only matched 7.x could not have been raised against a lockfile that has no 7.x in it.

## What actually pulled tar 6

`electron-rebuild@3.2.9` was the **sole** root:

```
codetracer@workspace:. -> electron-rebuild@3.2.9
  -> tar@^6.0.5
  -> node-gyp@9.4.1 -> tar@^6.1.2
     -> make-fetch-happen@10.2.1 -> cacache@16.1.3 -> tar@^6.1.11
```

`electron-rebuild` is the deprecated pre-transfer name; `@electron/rebuild@4.2.0` is the maintained continuation. It depends on `node-gyp@^12.2.0` (which uses `tar@^7`), and still ships the same `electron-rebuild` binary, so the `node_modules/.bin/electron-rebuild` path named in `docs/walkthrough.md` keeps working. No build script invokes it — the only other references in the tree are comments.

## Proof the critical is closed — from the lockfile, not a green check

Every `tar` block in the resulting `node-packages/yarn.lock`:

```
"tar@npm:^7.4.3, tar@npm:^7.5.2, tar@npm:^7.5.4":
  version: 7.5.22
```

One entry, 7.5.22. `tar@6.2.1` is gone. 7.5.22 clears **all 12** open tar alerts on this manifest (ranges `<= 7.5.2`, `<= 7.5.3`, `< 7.5.7`, `< 7.5.8`, `<= 7.5.9`, `<= 7.5.10`, `<= 7.5.15`, `<= 7.5.16`, `<= 7.5.17` ×2, `<= 7.5.18` ← the critical, `<= 7.5.20`), not just the critical one.

## Delta audit

By **distinct (package, version) pair** across all `resolution:` lines — 1377 pairs before, 1316 after:

- **67 removed** — exactly the `electron-rebuild` / `node-gyp@9` / `cacache@16` subtree, including `tar@6.2.1`.
- **6 added** — `@electron/rebuild@4.2.0`, `node-gyp@12.4.0`, `node-abi@4.35.0`, `node-api-version@0.2.1`, `read-binary-file-arch@1.0.6`, `undici@6.28.1`.
- **zero downgrades.**

Of the 6 additions, `undici@6.28.1` is the only one named by *any* of the 194 open alerts. All 21 undici ranges are `< 6.28.0`, `< 6.27.0`, `< 6.24.0`, or `>= 7.0.0` — 6.28.1 (the highest published 6.x) matches none. So this introduces no new alert at any severity.

A name-keyed diff makes `glob` and `rimraf` look like downgrades. They are not: both were *removed* with the node-gyp 9 tree, and the surviving copies already existed at those versions.

## yarn-cache hashes

The lockfile change invalidates both pinned hashes. Per the procedure documented in `yarn-project.nix` itself, a stale-but-once-valid hash is **substituted from the binary cache and never rebuilt**, so no mismatch is reported — it surfaces later as `YN0056: Cache entry required but missing`. Both were therefore forced with `sentinelHash`.

- **`aarch64-darwin`: `sha512-rJ3Oka4DrMUZPVorCFF5usgDOBwQaXM/6AfzvVohcMiR7qB1EkYkr+hCNn3AVu+9Hl4TAZ472TBFhfSwxE8XJw==`** — measured locally; re-pinned and rebuilt, after which `yarn-cache.drv` no longer appears in the build plan.
- **`x86_64-linux`: still on `sentinelHash`.** It can only be measured on a Linux builder and none is configured here (`builders` empty, `extra-platforms = x86_64-darwin`). **The first CI run will fail its Nix jobs with a hash mismatch printing the true `got:` — that value must be pasted in before this merges.** Until then the Nix jobs are expected red, by construction.

`yarn-project.nix` was verified byte-identical after the yarn invocation (the nixify plugin was disabled), so the per-system block was not collapsed.

## Note on #716

#716 rewrites `node-packages/yarn.lock` (+856/−1079) but leaves `node-packages/yarn-project.nix` **byte-identical to `stable`**, which does carry the per-system hash block. Both of its hashes are therefore stale by exactly the mechanism above — that branch would not report a mismatch, it would fail with `YN0056`. Separately, #716's own `storybook/package-lock.json` still contains `tar@6.2.1`, so it does not close the tar issue there either.

### DIFF
--- a/node-packages/package.json
+++ b/node-packages/package.json
@@ -58,6 +58,7 @@
     "yamljs": "^0.3.0"
   },
   "devDependencies": {
+    "@electron/rebuild": "^4.2.0",
     "@eslint/js": "^9.39.1",
     "@playwright/test": "^1.56.1",
     "@testing-library/dom": "^10.4.1",
@@ -67,7 +68,6 @@
     "@typescript-eslint/eslint-plugin": "^8.47.0",
     "@typescript-eslint/parser": "^8.47.0",
     "css-loader": "7.1.2",
-    "electron-rebuild": "^3.2.9",
     "env-paths": "^3.0.0",
     "eslint": "^9.39.1",
     "eslint-config-love": "^133.0.0",
--- a/node-packages/yarn-project.nix
+++ b/node-packages/yarn-project.nix
@@ -90,11 +90,14 @@ let
     # procedure above -- the previous values were real, so Nix would have
     # substituted the stale caches from the binary cache and never run the
     # builder, surfacing as ``YN0056: Cache entry required but missing``.
-    # x86_64-linux is left on the sentinel deliberately: this machine is
-    # aarch64-darwin, so the Linux ``got:`` can only be measured on CI. The
-    # first CI run on this branch will fail its Nix jobs with a hash mismatch
-    # and print the true value; paste it here before merging.
-    "x86_64-linux" = sentinelHash;
+    # Measured on CI (run 34337422212) after the sentinel forced a real build of
+    # yarn-cache.drv on x86_64-linux -- no Linux builder is configured on the
+    # machine this branch was prepared on, so this value could not be obtained
+    # locally. Three independent jobs -- test-non-gui (nixos, eph-linux-x64-g1),
+    # reprobuild-linux-smoke, and ct-test release gate (M16 provider matrix) --
+    # each built /nix/store/jkmf1s93nsmg5n2hyxc9hggwyj044yqv-yarn-cache.drv and
+    # reported this identical ``got:`` value.
+    "x86_64-linux" = "sha512-YOh4Lqjn3A2V2gYD+sspl5ZVOZJ+SCRPLHwp+DD03wq9t4eHQv2kXEv6aqhYW4ybT5itznWOe6E1SxfFDC57sg==";
     # Measured locally on aarch64-darwin: the sentinel above forced a real build
     # of yarn-cache.drv (no substitute can exist for it), and the mismatch
     # reported this ``got:``. Re-pinned and rebuilt to confirm it is stable.
--- a/node-packages/yarn-project.nix
+++ b/node-packages/yarn-project.nix
@@ -82,15 +82,23 @@ let
     "sha512-AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA==";
 
   cacheOutputHashes = {
-    # Measured on CI (run 34233823494) after the sentinel above forced a real
-    # build of yarn-cache.drv on x86_64-linux. Six jobs on eph-linux-x64-g1 --
-    # ViewModel headless tests, test-non-gui (nixos), cross-process
-    # value-origin, ct-test cross-language provider gate, ct-test release gate
-    # and reprobuild-linux-smoke -- each built
-    # /nix/store/lciwdnrgfbdq3k7py1w2n0k58w35svfd-yarn-cache.drv and reported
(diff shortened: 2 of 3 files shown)


zah:
## CI verdict: no regressions, measured against a PR-event control

Both hashes are now confirmed **on CI**, not just locally:

- **`reprobuild-linux-smoke`: green.** It failed on the previous run (34337422212) *solely* on the sentinel hash mismatch, and passes now — that is the x86_64-linux hash verified.
- **`reprobuild-macos-smoke`: green.** This is the real confirmation of the aarch64-darwin hash. My local check was **not** sufficient evidence and I want to be explicit about why: a fixed-output path is addressed by its *declared* hash, so once my first local build had realised that path, any later local build simply reused it — it never re-derived the content. CI's runner had no such path and built it for real.
- **`dmg-build`, `dev-build`, `appimage-build`, `nix-build`: all green** — the jobs that actually consume `node-modules-derivation`, i.e. the ones that would break if dropping `electron-rebuild` (and with it `lzma-native`) broke packaging.

### Non-regression comparison

Control is #738 — an independent PR whose only change is one trailing newline, branched from the **same base commit** (`ceb111a13`), so it is a `pull_request`-event run with the same job set. Populations: 46 jobs on each run.

| | PR #737 | control #738 |
|---|---|---|
| failures | 12 | 14 |
| **PR-only failures** | **0** | — |
| control-only failures | — | 2 |

The 2 control-only failures are `origin-DAP (materialized Python, Windows)` and `windows-rust-components`, both with **steps == 0** — no code ran; the `eph-win-x64` allocation loss, not a real result.

Failing-job-name overlap is not sufficient evidence, so **causes were compared for every job red on both sides**:

- 9 of 12 have byte-identical extracted error signatures.
- `windows-named-pipe-tests` — both fail at the same PowerShell step ("Persist the compiler/linker search paths"), logs 42853 vs 41974 bytes.
- `test-ui-tests (nixos, eph-linux-x64-g2)` — both fail the same 23 specs on the same stale-built-stylesheet guard; only the incidental `.styl` filename named in the message differs.
- `queue-watchdog` — both exit 4 (nothing servicing the run; the pool is saturated).

**One job's cause genuinely differs**, and it is worth stating plainly rather than letting it cancel out of a set difference: `windows-bootstrap-smoke` fails on both, but on #737 with **steps == 0** (log is still the placeholder — no code of ours ran) while on the control it is a real relocatability-check failure. A change to `node-packages/` cannot cause a Windows runner to fail to start, so this is not a regression — but it is *not* the same failure as the control's.

### Not yet measured

5× `appimage-cross-distro` and (on #737) `push-tag` are still queued on both runs — the `eph-linux-x64` pool is saturated. They are pending **identically on both sides**, so they are simply unmeasured rather than divergent. `appimage-build` itself, which is the part this change could plausibly affect, is green.

PR #738, opened 2026-09-09
ci: no-op control for #737 non-regression comparison (NOT FOR MERGE)

zah:
Control PR. **Do not merge.**

Its only change is one trailing newline in `docs/walkthrough.md`. It exists so that #737's CI failures can be compared against a `pull_request`-event run off the **same base commit** (`ceb111a13`), rather than against a `push` run — `pull_request`-gated jobs are absent from push runs and would read as new failures.

Close once #737 is resolved.


zah:
Throwaway PR-event control for #737's non-regression comparison. Closing; it was never intended to merge.

PR #739, opened 2026-09-09
feat(watchpoints): implement setDataBreakpoints — the mock was the only implementation

zah:
> **STACKED ON #730.** Base is `fix/py-add-watchpoint-verdict`, not `dev`, because #730 is still open. Merge #730 first; this PR's diff then reads as its own commit alone. It builds on #730 and undoes none of it — `Unobserved(UnobservedReason)` survives and gains a caller.

## The premise, established before building

**There is no GUI surface for watchpoints, and this PR does not invent one.** The Nim frontend has no watchpoint affordance anywhere: no menu item, no context action, no pane, no keybinding, no Karax component. Its only two `.nim` hits are a dead `RRGDBError`/`RRGDBStatus` enum member (zero use sites outside its own declaration) and an unimported 53-line DAP response sketch in `src/frontend/ui/dap/response.nim`. The GUI's "watch expression" feature in the State pane is a re-evaluated expression, not a data breakpoint.

The **user-reachable** surface is real, though, and it is what made this worth building:

| Surface | Where |
| --- | --- |
| Python API `Trace.add_watchpoint` | `python-api/codetracer/trace.py:750` — the product's public API |
| MCP `exec_script` tool description | `src/backend-manager/src/mcp_server.rs:98` — advertises watchpoints to every agent that lists the tool |
| MCP `trace_query_api` prompt | `mcp_server.rs:507` |
| The book | `docs/book/src/reference/mcp-tools.md:39` claims "breakpoints and watchpoints" |

All four told users watchpoints worked. None of them did.

## What was broken

`ct/py-add-watchpoint` answered `success: true` with a `watchpointId` before the backend saw the request. #730 made it wait. What the wait exposed was **not** `verified: false` — `db-backend`'s DAP dispatch had **no `setDataBreakpoints` arm at all**. The command fell through to `dap_command_to_step_action` and came back as `command setDataBreakpoints not supported here`: the `_` arm meant for commands nobody had thought about, indistinguishable from a typo in the command name.

**Why nothing caught it:** the only implementation of `setDataBreakpoints` in the tree was the daemon's **mock**, answering `verified: true` unconditionally. A test double more capable than the component it stands in for is a blindfold — every test of the watchpoint path passed against a backend that could not do the thing being tested.

## What a watchpoint is, over a recording

A live debugger uses a hardware watch register. A recording has no CPU — it has the value each variable held at each step. So: **a value-change watchpoint.** `continue` stops at the first later step where the named variable's recorded value differs from what it held before.

Two rules keep it honest, both asserted: coming into scope is not a change (first sighting sets the baseline silently — otherwise it stops on entry to every call that declares the name), and a step that doesn't record the variable leaves the baseline alone (out of scope ≠ changed to nothing).

## What the replay model genuinely cannot support

A **closed set**, `DataBreakpointRefusal` (6201–6206, continuing `OriginErrorCode`'s 61xx convention). The code rides in the response body as `refusalCode` plus a stable `refusal` token, so callers branch on an integer instead of parsing English — which is the whole difference from the free-text fallthrough it replaces.

| Code | Token | Why |
| --- | --- | --- |
| 6203 | `accessTypeNotRecorded` | **The one that is intrinsic, not merely unimplemented.** A recording samples what a variable *held*, never that a value was *read*. Reading `x` changes nothing about `x`, so no scan recovers it. `write` is the only access type a value table can answer. |
| 6202 | `expressionNotAWatchableVariable` | The recording indexes by variable, not address or expression — `obj.field`, `arr[0]`, `*ptr`, `counter + 1` are unwatchable. |
| 6204 | `conditionNotSupported` | Refused rather than silently ignored: a watchpoint that drops its condition stops in the wrong place with nothing to say so. |
| 6205 | `variableNotInTrace` | Well-formed name, not one this recording captured. Separate from 6202 because the user's fix differs. |
| 6206 | `backendLacksValueHistory` | MCR/emulator and recreator keep no per-step value table — they already refuse `load_history` for the same reason. |
| 6201 | `emptyDataId` | A watchpoint has to name something. |

Per-entry verdicts in request order; a refusable entry doesn't fail the request; a refused entry is never installed.

## The mock/real reconciliation

The rules now live in `libs/ct-data-breakpoints`, a leaf crate **both** the real backend and the mock link against. The mock can't over-claim because it isn't the mock that decides — it supplies only its own *vocabulary* (the 5 variables its fake trace actually reports).

`conformance_cases()` is a shared table both sides are driven through, from their own crates:
- db-backend → `conformance_cases_agree_with_the_real_backend`
- backend-manager → `mock_matches_the_shared_data_breakpoint_verdict` (drives the mock's real code path, not a re-implementation)

**Mutation-checked.** Reverting the mock to unconditional `verified: true` turns **3 of 5** parity tests red:
```
the mock answered verified:true for DataBreakpointRequest { data_id: "counter",
access_type: Some("read"), .. }, which the shared rules REFUSE with
accessTypeNotRecorded. This is the exact defect this module exists to catch.
```

## Red before, green after

`src/db-backend/tests/dap_data_breakpoints_test.rs` — **10 tests, all 10 observed RED**, every one with:
```
dispatch of `setDataBreakpoints` failed: command setDataBreakpoints not supported here
test result: FAILED. 0 passed; 10 failed
```
All 10 pass after. They cover dispatch, the stop at the first change (step 3 — *not* step 1, which would mean firing on any step recording the variable, and *not* the last step, which is what users saw as `StopIteration`), advancing to the next change rather than re-firing, a constant never firing, replace-semantics clearing, per-entry closed-set refusals, and a refused entry not being installed.

`test_real_py_add_watchpoint_does_not_answer_for_the_backend` asserted watchpoints *fail*; #730 flagged its first assertion as the thing to update deliberately when the feature landed. Renamed to `..._reports_the_backends_own_verdict` and inverted against `RUBY_TEST_PROGRAM` (locals `a`, `result`, `x`, `y`): `counter` refuses **by name** (`variableNotInTrace`, not the fallthrough), `x` is **accepted with a positive id**.

## A defect this activates, fixed here

`ct/py-remove-watchpoint` re-sent `setDataBreakpoints` fire-and-forget under `UnobservedReason::SetPointsResyncAfterRemoval`, excused by "a watchpoint that was never set cannot fail to be removed either, so no client is currently misled". That excuse held **only while the backend refused every watchpoint**. It expired the moment they started binding.

The removal now waits, under its own `PendingPyRequestKind::RemoveWatchpoint`. On refusal it restores the watchpoint under the same id and answers `success: false` instead of a `removed: true` that isn't true. A reason recorded in `UnobservedReason` is a claim about the world; when the world changes, the claim has to be rechecked.

## Surfaces now tell the truth

`dataBreakpointInfo` is implemented (it's the handshake a DAP client uses to decide whether to *offer* the affordance, and it reports `accessTypes: ["write"]` so a conforming client never offers a read watchpoint that can only be refused). `initialize` advertises `supportsDataBreakpoints: true`. The refusal reaches the user where they asked — `TraceError.refusal` / `.refusal_code`. New `docs/book/src/reference/watchpoints.md` (mirrored into book-isonim) documents semantics and the full refusal table. The MCP tool description and prompt, the Python docstrings and `.pyi`, `ct-dap.md`, and the bench docs all state the real limits. `codetracer-bench`'s `Operation::Watchpoint` now emits a real `setDataBreakpoints` instead of `None`.

## CI coverage

`libs/ct-data-breakpoints` is wired into `just test-rust`. Without that line its tests are **dark** — a plain path dependency of two crates, neither of which builds a dependency's test targets — and a dark conformance table is not an anchor for the parity tests hanging off it. `ci/test/rust-test-crate-coverage.sh` (added on this base) caught exactly that; it now passes with the recorded 5-entry baseline unchanged.

## Executed locally — macOS aarch64

| Suite | Result |
| --- | --- |
| `libs/ct-data-breakpoints` | 6/6 pass |
| `db-backend --lib` | 648 pass, 2 fail — **identical to baseline** (`tracepoint_interpreter::{array_indexing,log_array}`, SIGSEGV recording a fixture, unrelated) |
| `db-backend dap_data_breakpoints_test` | 10/10 pass (**10/10 red before**) |
| `db-backend` dap_protocol, dap_column_breakpoint{,_with_condition}, dap_column_tracepoint | all pass |
| `backend-manager --bins` | 216 pass, 1 fail — the pre-existing `browser_stream_host::verify_reframing_...` needing `scripts/materialize-recording.sh`. Baseline 209/1; +7 new tests |
| `ci/test/rust-test-crate-coverage.sh` | OK, baseline unchanged |
| clippy (db-backend lib+tests, shared crate) | no warnings from this change |

**NOT executed locally:** `test_real_py_add_watchpoint_reports_the_backends_own_verdict` self-skips — it needs a built `db-backend` binary and the Ruby recorder.

**Correction to #730's stated limit:** this checkout *can* build the db-backend library, via `--no-default-features --features io-transport,syntax-highlight` (skipping `nim-reader`, which is the feature that needs the nimble-provisioned Nim toolchain). That is how all 10 dispatch tests were run and how the red-before evidence was captured. #730 reported the crate could not be built here at all.

---

## Follow-up commits on this branch

Three test-only commits after the implementation, each pinning behaviour the implementation created:

- **`reverse Continue`** — the mirror of forward: from the last step (`counter == 9`) the backward walk passes steps 6 and 5 (both 9) and stops at step 4, the first step going backwards holding a different value. Pinned so the symmetry is deliberate rather than whatever the scan happened to do. Its first draft also found a trap now recorded in a comment: `handler.step_id` and the replay engine's position are **separate**, and the scan seeds its baseline from the *engine's*. Assigning the handler field alone leaves the engine at step 0 and the walk runs off the front of the trace. The forward tests never noticed because both positions start at 0.
- **The flow test double** — `ReplaySession::has_per_step_values` defaults to `false`, so a backend must *opt in* to claiming it can answer watchpoints. Only `MaterializedReplaySession` does. `flow_preloader`'s `MockReplay` — the tree's *other* test double — is pinned as not over-claiming. A capability that defaults to "yes" makes the mock-over-claims mistake silent and free; one that defaults to "no" makes it a visible line someone has to write.
- **The refusal code reaching the caller** — plus the additive case: a backend supplying no code must still round-trip its message. An absent code must not become an absent *message*.

## CI

`mcr-dap-flow` is red here and is **not** caused by this change. Cause, from the job log:

```
bash: line 1: just: command not found
error: Recipe `ensure-ct-native-replay` failed with exit code 127
```

Exit 127 — nothing compiled, no test ran. Confirmed **by cause, not by name** against an unrelated PR's run the same day (`retire-legacy-json-sidecar-readers`, run 34316581473), whose log carries those three lines byte for byte. The workflow's last 20 runs are *all* failures, across `dev` pushes, `main`, and many unrelated PRs, going back to 2026-08-30.

The control used throughout is **PR #730** — a `pull_request` event on the same base, one commit apart — which shows **13 failures**, matching the known ~12-job `dev` baseline. `lint-rust`, `lint-bash`, `lint-nim`, `lint-nix`, `nix-build` and `dev-build` are all green on that control, so they are the jobs that would actually indict this change.


zah:
## CI triage — compared by **cause**, against a `pull_request`-event control

Control throughout: **#730** (`pull_request`, same base tree, one commit apart). It shows **13 failures**, matching the known ~12-job `dev` baseline.

### The one thing that is genuinely about this PR

**`workspace-lock-freshness`** — SUCCESS on the control, FAILURE here. **It is caused by the *stacking*, not by the code.** The check keys off the PR's **base** branch, and its own output says so:

```
NO PUBLISHER IS CONFIGURED FOR BRANCH 'fix/py-add-watchpoint-verdict'
Branch under test: fix/py-add-watchpoint-verdict
CAUSE: the branch, not the workspace. 'fix/py-add-watchpoint-verdict' is NOT in the
trigger lists of publish-workspace-lock.yml, which publishes for: cloud dev stable
— so no lock was ever going to exist for this commit, or for any other
commit on this branch.
Nothing done in a developer's workspace can change that.
```

The control passes it only because *its* base is `dev`. **This goes green by itself once #730 merges and this PR is retargeted to `dev`** — no code change can affect it.

### Positive signal: the new crate is no longer dark

`test-non-gui (macos)` shows the `just test-rust` line I added actually running in CI:

```
Compiling ct-data-breakpoints v0.1.0 (.../libs/ct-data-breakpoints)
    Starting 6 tests across 1 binary
        PASS (1/6) ct-data-breakpoints tests::every_conformance_case_matches_the_shared_verdict
        PASS (2/6) ct-data-breakpoints tests::refusal_codes_are_distinct_and_in_the_62xx_block
        PASS (3/6) ct-data-breakpoints tests::the_conformance_table_exercises_every_refusal_variant
        PASS (4/6) ct-data-breakpoints tests::a_read_watchpoint_is_refused_for_the_reason_that_is_intrinsic
        PASS (5/6) ct-data-breakpoints tests::identifier_rules
        PASS (6/6) ct-data-breakpoints tests::the_conformance_table_has_accepted_cases_too
     Summary 6 tests run: 6 passed, 0 skipped
```

The control has **zero** references to that crate — it predates the wiring. The shared conformance table that anchors both parity suites is now executed in CI.

### The other four failures — all pre-existing, verified by cause

| Job | Cause | Evidence it is not this PR |
| --- | --- | --- |
| `test-non-gui (macos)` | `direnv: error .../codetracer-native-recorder/.envrc is blocked` → `build_native_api.sh exited with status 1`, panic at `build.rs:972`. Exit 101 at **build**, no test ran. | Unrelated PR `retire-legacy-json-sidecar-readers` hits the identical block on a different runner, also with zero tests run. Runner-dependent: on runners where the sibling is provisioned (`ci/reattach-lane-reachability`, `test/trace-origin-renderers`, `fix/inline-run-test-claimed-start`, and the control) every branch gets the **same** `617 tests run: 615 passed, 2 failed` — the two `tracepoint_interpreter::{array_indexing,log_array}` failures that also fail on a clean local checkout. |
| `test-ui-tests (macos)` | Fails at `Build ct-mcr via codetracer-native-recorder (macOS)`; all test steps **skipped**; then the same two Playwright specs. | **Step-for-step identical** to the control, including which steps skip, and the `##[error]` lines match exactly: `event_log_is_static_across_jumps.spec.ts:152` and `:212`, exit codes 1 and 101. |
| `mcr-dap-flow` | `bash: line 1: just: command not found` → `Recipe ensure-ct-native-replay failed with exit code 127`. Nothing compiled. | Byte-identical in an unrelated PR's run the same day (run `34316581473`). The workflow's **last 20 runs are all red**, across `dev` pushes, `main`, and unrelated PRs, since 2026-08-30. |
| `visual-replay-regression-gate` | In the control's 13-failure set. | Pre-existing. |

All three environmental failures share one root cause: the **`codetracer-native-recorder` sibling is unbuildable on these macOS runners** — which is also the step that fails on the control's own `test-ui-tests`.

### Still pending

`lint-rust`, `lint-bash`, `nix-build`, and `test-non-gui (nixos)` are queued. These are the jobs that would actually indict this change (all green on the control). Their gates were run locally first:

- `backend-manager: cargo clippy -- -D warnings` → pass
- `db-backend: cargo check -D warnings` → pass
- `ci/test/rust-test-crate-coverage.sh` (the gate inside `lint-bash`) → **OK, baseline unchanged** — this is the gate that caught the new crate being dark, and the reason the `just test-rust` line exists.

zah:
## CI triage, corrected and consolidated

My earlier comment called `workspace-lock-freshness` "the one thing genuinely about this PR". **That was incomplete** — there are **three** red jobs caused by the *base*, and none by the watchpoint code. Correcting it here rather than leaving it to be re-derived.

The base is `fix/py-add-watchpoint-verdict` (#730's branch), which is **21 commits behind `origin/dev`** and has **no published workspace lock**. Those two facts explain all three.

### A. Caused by the stacked base — clears on rebase/retarget, no code change possible

| Job | Cause |
| --- | --- |
| `workspace-lock-freshness` | The check names it: `NO PUBLISHER IS CONFIGURED FOR BRANCH 'fix/py-add-watchpoint-verdict'` … `CAUSE: the branch, not the workspace` … `publishes for: cloud dev stable` … *"Nothing done in a developer's workspace can change that."* |
| `js / codetracer-desktop@…` | Hard error at *Setup dev env + clone the remaining siblings*: `No workspace lock for codetracer (candidates: 4a0b843ae…)` — that SHA is the base. Every build/test step after it **skipped**. Zero references to anything this PR touches. |
| `lint-bash` | `FAIL a rebuilt renderer makes the copied corpus stale / the command unexpectedly succeeded / output: FRESH`. **Already fixed on `dev`** by `a4681be78` *"ci: state the storybook fixture's mtimes instead of racing for them"*, whose message quotes this exact failure text and explains it: `requireFreshStorybookStatic` compares strictly, and two adjacent bare `touch`es can land in the same clock tick. That commit postdates this base. This PR touches no storybook, renderer, stylesheet or component file. |

`lint-bash` is otherwise **42 of 43 steps OK**, and it contains the gate that matters here:

```
dark crates:            5
OK: the dark-crate set matches the recorded baseline (5 entr(y/ies))
```

— i.e. the new `libs/ct-data-breakpoints` crate is **not** dark. That gate is the reason the `just test-rust` line exists.

### B. Pre-existing, each verified by cause against a control

| Job | Cause | Control evidence |
| --- | --- | --- |
| `rr-backend-tests` | `==> FAILED: nim-flow rust-flow`, `should find at least some of the expected variables` | **Identical on `dev` itself** (push run, job `101410759931`) and on unrelated PR `retire-legacy-json-sidecar-readers`. Every run of this workflow is red, including 9 consecutive `dev` pushes. Zero watchpoint references in the log. |
| `test-non-gui (macos)` | `direnv: … .envrc is blocked` → `build_native_api.sh exited with status 1`, panic at `build.rs:972`. Exit 101 at **build**; no test ran. | Same block on unrelated PR `retire-legacy-json-sidecar-readers`. On runners where the sibling *is* provisioned, every branch — including the control — gets the same `617 tests run: 615 passed, 2 failed`. |
| `test-ui-tests (macos)` | Fails at `Build ct-mcr via codetracer-native-recorder`; test steps skipped. | **Step-for-step identical** to the control, and the `##[error]` lines match exactly (`event_log_is_static_across_jumps.spec.ts:152` and `:212`). |
| `mcr-dap-flow` | `just: command not found` → exit 127. Nothing compiled. | Byte-identical on an unrelated PR the same day. Last **20** runs all red since 2026-08-30. |
| `visual-replay-regression-gate` | In the control's 13-failure set. | Pre-existing. |

### Positive signal

`test-non-gui (macos)` shows the new crate compiling and running **in CI**, by name:

```
Compiling ct-data-breakpoints v0.1.0 (.../libs/ct-data-breakpoints)
    Starting 6 tests across 1 binary
        PASS (1/6) tests::every_conformance_case_matches_the_shared_verdict
        PASS (2/6) tests::refusal_codes_are_distinct_and_in_the_62xx_block
        PASS (3/6) tests::the_conformance_table_exercises_every_refusal_variant
        PASS (4/6) tests::a_read_watchpoint_is_refused_for_the_reason_that_is_intrinsic
        PASS (5/6) tests::identifier_rules
        PASS (6/6) tests::the_conformance_table_has_accepted_cases_too
     Summary 6 tests run: 6 passed, 0 skipped
```

The control has **zero** references to that crate — it predates the wiring.

### Recommendation

Merge #730, then retarget this to `dev` (or rebase the stack onto current `dev`). That clears **A** in full. I have deliberately not rebased #730's commit myself — it is someone else's open PR and merge sequencing is yours.

zah:
## `test-non-gui (nixos)` — and an honest limitation

**Identical to the control, at the same line.** Both this PR and #730 fail at:

```
thread 'main' panicked at build.rs:170:5:
db-backend requires the sibling `codetracer-native-recorder` checkout, and it was not found.
...
error: Recipe `test-rust` failed with exit code 101
```

Same `build.rs:170:5`, same exit code, same recipe. The sibling is not cloned on that runner.

**The new crate passed here too**, before the build died — so it is confirmed running in CI on *both* legs:

```
Compiling ct-data-breakpoints v0.1.0 (.../libs/ct-data-breakpoints)
    PASS (1/6) tests::a_read_watchpoint_is_refused_for_the_reason_that_is_intrinsic
    PASS (2/6) tests::the_conformance_table_has_accepted_cases_too
    PASS (3/6) tests::refusal_codes_are_distinct_and_in_the_62xx_block
    PASS (4/6) tests::identifier_rules
    PASS (5/6) tests::every_conformance_case_matches_the_shared_verdict
    PASS (6/6) tests::the_conformance_table_exercises_every_refusal_variant
 Summary 6 tests run: 6 passed, 0 skipped
```

### The limitation, stated plainly

Because `replay-server` fails to **build** on both `test-non-gui` legs — the Linux one identically to the control, the macOS one from the `direnv`-blocked sibling — **`src/db-backend/tests/dap_data_breakpoints_test.rs` has not been observed running in CI.** Neither has the `backend-manager` parity suite on those legs.

What has actually been observed, and where:

| Evidence | Where |
| --- | --- |
| `ct-data-breakpoints` 6/6 | **CI**, both macOS and Linux |
| dark-crate gate `OK`, baseline 5 unchanged | **CI** (`lint-bash`) |
| `dap_data_breakpoints_test` 11/11 — and **10/10 red** before the fix | **local only** (macOS, `--no-default-features --features io-transport,syntax-highlight`) |
| `backend-manager` 218 pass / 1 pre-existing fail, incl. 5 parity + 4 refusal tests | **local only** |
| mock mutation → 3 of 5 parity tests red | **local only** |
| `cargo clippy -- -D warnings`, `cargo check -D warnings` | **local only** (`lint-rust` still queued) |

Note that even on runners where the sibling *is* provisioned, the `617 tests` figure is `nextest --bin replay-server` — the binary's unit tests. The integration targets come from the separate `--test '*'` line, which is downstream of the same build.

This is not a gap this PR introduces — the db-backend build has been failing on these runners for the control too — but it does mean the strongest evidence for the watchpoint implementation is currently local, and I would rather say so than let the green boxes imply otherwise.

zah:
## Final triage tally — 10 red, 0 attributable to the watchpoint code

Every failure below was checked **by cause**, against a `pull_request`-event control (#730, same base branch) or against `dev` itself. `lint-rust` / `lint-nim` / `lint-nix` are still queued.

### A — artifacts of the stacked base (3)

Base is `fix/py-add-watchpoint-verdict`: **21 commits behind `dev`**, **no published workspace lock**. Nothing in the diff can affect these.

| Job | Cause |
| --- | --- |
| `workspace-lock-freshness` | The check names it: *"CAUSE: the branch, not the workspace… publishes for: cloud dev stable… Nothing done in a developer's workspace can change that."* |
| `js / codetracer-desktop@…` | Hard error `No workspace lock for codetracer (candidates: 4a0b843ae…)` — that SHA **is** the base. All build/test steps skipped. |
| `lint-bash` | Storybook mtime race, **already fixed on `dev`** by `a4681be78`, whose message quotes this failure text verbatim. Otherwise **42/43 steps OK** — including the dark-crate gate. |

### B — pre-existing, cause matched (7)

| Job | Cause | Same cause seen on |
| --- | --- | --- |
| `beam-flow-dap` | `beam-flow-cross-repo.sh: line 88: cargo: command not found` (127) | **`dev` push** + unrelated PR, same line number. Last 12 runs all red. |
| `mcr-dap-flow` | `just: command not found` (127) | Unrelated PR, byte-identical. Last 20 runs all red since 2026-08-30. |
| `rr-backend-tests` | `==> FAILED: nim-flow rust-flow` | **`dev` push** + unrelated PR, identical assertion. Every run of the workflow red. |
| `test-non-gui (nixos)` | `panicked at build.rs:170:5: db-backend requires the sibling codetracer-native-recorder checkout` | **The control, at the identical line and exit code.** |
| `test-non-gui (macos)` | `direnv: … .envrc is blocked` → `build.rs:972` | Unrelated PR. Where the sibling *is* provisioned, every branch incl. the control gets the same `617 run / 615 pass / 2 fail`. |
| `test-ui-tests (macos)` | Fails at `Build ct-mcr via codetracer-native-recorder` | **Step-for-step identical** to the control; `##[error]` lines match exactly. |
| `visual-replay-regression-gate` | In the control's 13-failure set | Pre-existing. |

Six of the seven trace to one root: **`codetracer-native-recorder` / dev-shell PATH is broken on these runners**, which is also the step that fails on the control's own `test-ui-tests`.

### Green, and load-bearing

`ci-contract-suites`, `docs-scope`, `origin-DAP (materialized Python, macOS)`, `reprobuild-{linux,macos}-smoke`, `shell-recorder-tests`, `ct-test release gate`.

Plus, executed in CI on **both** legs: `ct-data-breakpoints` **6/6 by name**, and `lint-bash`'s gate reporting `dark crates: 5 — matches the recorded baseline`, i.e. the new crate is wired and no longer invisible.

The honest limitation from the previous comment stands: because `replay-server` fails to *build* on both `test-non-gui` legs (identically to the control on Linux), **`dap_data_breakpoints_test` has not been observed running in CI** — its 11/11 green and 10/10 red-before are local.

zah:
### Addendum — 11th failure triaged: `cross-process value-origin (Linux)`

Pre-existing, in the control's set, and **cause matched byte-for-byte**. Same failing step on both this PR and #730 — *Build frontend (needed by gated Playwright stage)* — ending in:

```
Initializing .tup in .../src
.tup repository initialized: .tup/db
```

The only mentions of this PR's code anywhere in that log are two successful dependency compiles —

```
Compiling ct-data-breakpoints v0.1.0 (.../libs/ct-data-breakpoints)
```

— i.e. the new crate builds cleanly as a dependency of the frontend build; the failure is downstream in `tup`/`build-once` and identical on the control.

**Running tally: 11 red, 0 attributable to the watchpoint code** — 3 stacked-base artifacts, 8 pre-existing with matched causes. `lint-rust`, `lint-nim`, `lint-nix` remain queued; their gates (`cargo clippy -- -D warnings`, `cargo check -D warnings`) were run locally and pass.

PR #756, opened 2026-09-15
fix(flake): mirror the org's current reprobuild pin so the build recipe compiles

zah:
## What was broken

`repro build` in this repo did not get as far as building anything: the build
recipe itself failed to compile.

```
repro.nim(1766, 40) Error: type mismatch
  [7] extraEnv = GateEnv: extraEnv: array[0..0, (string, string)]
  unknown named parameter: extraEnv
```

The ten gate actions in `repro.nim` declare `PYTHONHASHSEED` on the action that
reads it, via `shell(extraEnv = ...)`. That parameter did not exist on the
`reprobuild` revision this repo pinned (`3a7ef0d7`, 2026-09-10) — it landed
two days later — so the recipe used a DSL the pinned `repro` could not parse.
The pin was the stale thing, not the recipe.

It survived because nothing in CI compiles `repro.nim` on a pull request:
`reprobuild-linux-smoke` builds a scratch project in a `mktemp -d`,
`windows-installer-build` is gated to `main`, and `test-ui-tests (macos)` —
the one job that does compile it — was already failing at exactly this error.

## The fix

- `reprobuild` → `69bb0a7d`, the revision `metacraft-labs/nixos-modules`
  publishes as the org-wide `repro`. `flake.nix` already names nixos-modules as
  the single source of truth for this SHA; it had simply not been mirrored.
  Neither the minimum viable revision nor `dev`'s tip: this is the one the rest
  of the org's shells are built from.
- `runquota` → `093f520d`, in lockstep, because
  `inputs.runquota-src.follows = "runquota"` means reprobuild is *compiled*
  against whatever this resolves to. This is the equality
  `scripts/test-flake-pin-alignment.sh` exists to enforce.
- Both lock nodes rewritten with `nix flake update reprobuild runquota`, never
  by hand, so `lastModified`/`narHash`/`rev` stay coherent.
- The comment next to the input now records that the pin is a *floor* as well
  as a mirror, and how to exercise it locally (a workspace checkout's
  `NIX_FLAKE_OVERRIDE_AUTO=1` silently overrides this input with the sibling
  `../reprobuild`, so a plain local `repro build` does **not** test the pin).

## Verification

Built against the pin itself, with the sibling override removed, confirming by
store path that the `repro` in play was the pinned revision:

- `repro build` → exit 0, recipe compiled (`providerInvocations: 1`), 30/30
  actions succeeded.
- `repro build gate-flake-pin-alignment gate-python-version-alignment
  gate-require-runtime-assets` → exit 0; all three declare their environment,
  which is the behaviour `extraEnv` was reached for.
- `just test-flake-pin-alignment` → OK (equality holds).
- `just test-flake-lock-node-dates` → OK, all 41 comparable nodes.
- `ci/test/flake-lock-metadata-test.sh` → 2 passed; all 31 direct github
  inputs record GitHub's own commit date.
- `just test-build-alignment` → 72 checks, 0 failures.
- `just test-reprobuild-linux-smoke` → passed.

`dev` has failures unrelated to this change (`beam-flow-dap` and the `beam`
arm, `ci-contract-suites`); `test-ui-tests (macos)` is the lane this fixes.


zah:
## Red-lane triage (neither is caused by this change)

### `ci-contract-suites` — pre-existing, byte-identical to `dev`

Same assertion block as `dev`'s last completed run, timestamps stripped:

```
  FAIL M1-M13: all thirteen mutations of real wiring are rejected
         M11 SURVIVED: a recorder added to a fan-out matrix but absent from the clone-list was not detected
1 of 18 assertions FAILED
```

`diff` over the whole assertion body against job `103886467315` (`dev`) is empty.
It is a CI clone-list mutation suite and reads no flake input.

### `test-non-gui (macos)` — runner assignment, not the diff

Two lanes red, and the base commit's own run (`dev` @ `7abb3a35`, job
`104285133782`) is red in this job too — differently, because it landed on a
different runner:

| | this PR (`mcl-006`) | `dev` @ `7abb3a35` (`mcl-004`) |
|---|---|---|
| `test-rust` | exit 101 — `codetracer-native-recorder/.envrc is blocked. Run 'direnv allow'`, so `build_native_api.sh` cannot run | exit 100 — `dap_handler::tests::test_js_load_locals_function_scope` |
| `test-sibling-backend-path` | FAIL (case A-negative) | PASS |

Both differences trace to whether the runner has the `codetracer-native-backend`
/ `codetracer-native-recorder` siblings checked out, not to anything in this
diff.

`test-sibling-backend-path` is worth writing down, because the lane is **not
hermetic in its A-negative leg** and that is a latent defect independent of this
PR:

* `probe_detect()` scrubs `CODETRACER_RR_BACKEND_PATH` /
  `CODETRACER_RR_BACKEND_PRESENT` with `env -u`, but **not `PATH`**;
* `scripts/detect-siblings.sh:756` derives presence from `PATH` alone —
  `command -v ct-native-replay &>/dev/null && export CODETRACER_RR_BACKEND_PRESENT=1`;
* so on a runner where the sibling *is* present and built (this PR: case B ran
  and passed), `ct-native-replay` is on `PATH`, the synthetic
  no-sibling workspace still reports `_PRESENT=1` with `_PATH` unset, and
  `assert_no_half_configured` fires.

On `mcl-004` case B reported `SKIP — no codetracer-native-backend checkout`, so
the leg passed for the opposite reason. The lane therefore passes only on a
runner that lacks the sibling. Not fixed here — different script, different
concern.

PR #763, opened 2026-09-21
Restore repro.nim: drop the two ctNimJs arguments the template no longer accepts

zah:
## The break

`repro.nim` on `dev` does not compile:

```
repro.nim(832, 31) Error: type mismatch
Expression: ctNimJs(definesValue = CommonNimDefines & HmrRendererDefines, ...
                    hotCodeReloadingOnValue = false)
Expected one of (first mismatch at [position]):
[5] template ctNimJs(definesValue: seq[string]; outputPath, sourcePath: string;
                 extraInputsValue: openArray[string] = [];
                 extraOutputsValue: openArray[string] = [];
                 debugInfoOnValue = false; sourcemapOnValue = false): BuildActionDef
  unknown named parameter: hotCodeReloadingOnValue
```

`ca1023df0` ("stop building the JavaScript bundles with Nim hot code
reloading") removed `hotCodeReloadingOnValue` from the local `ctNimJs`
template and stripped it from the call sites that passed `= true` — but left
it on the two that passed `= false`: `frontend-ui-js` (837) and
`frontend-subwindow-js` (881).

## The fix

Delete both arguments. Nothing else changes: the flag was already off, the
template already refuses it, and the template's own NOTE records why it must
stay absent — so the per-call-site restatement at the subwindow site goes
with it rather than describing a parameter that no longer exists.

## Why it went unnoticed

Nothing that compiled this file was reaching it. Reprobuild's CI `Test` job
builds `ct` through this recipe, but its sibling-clone step was failing
earlier, on a token mint that asked GitHub for a repository under its
pre-rename name. With that fixed (metacraft-labs/reprobuild#355), the job now
clones the siblings, reaches `Build codetracer ct (via reprobuild)`, and dies
here.

That same job is the verifier for this change.


zah:
**The `reprobuild-linux-smoke` red here is not from this PR.** Recording the evidence so it is not mistaken for one.

The smoke build itself passed. The job failed on a later, unrelated step — "Verify test-reprobuild-hcr-mcr-dap fails loudly unsupported on Linux" (job `106504081929`):

```
repro build: error: unknown_target: no build target matches 'ct-mcr'
error: Recipe `ensure-ct-mcr` failed with exit code 2
```

`ensure-ct-mcr` (`justfile:5102`) has two arms: SKIP with `exit 0` when `CT_CODETRACER_NATIVE_RECORDER_SIBLING` is unset, otherwise `cd $sibling && repro build ct-mcr`. `codetracer-native-recorder`'s `repro.nim` declares **no build targets at all** — `ct-mcr` appears there only in two doc comments — so the second arm can never succeed.

On the `dev` run 3.5 h earlier (job `106427423681`) the same step was **green**, reaching the expected `UNSUPPORTED: … requires macOS arm64` line in 14 s, because the sibling was not detected and it took the SKIP arm. On this run the dev-shell reported `sibling: codetracer-native-recorder (repo present, ct_cli not built) detected`, which flipped the arm. Nothing in a two-line argument deletion can affect sibling detection.

So the step is green exactly when it does nothing and red the moment it runs. Filed separately; it wants either a real `ct-mcr` target in the recorder's recipe, or the `just build-ct-mcr` path the dev-shell warning already recommends.

**On verification of this PR itself:** I have no local compile proof yet — the recipe needs codetracer's sibling set and dev shell, which this branch's worktree does not have, and I stopped a long local attempt rather than starve an unrelated full-suite measurement running on the same host. The compiler's own diagnostic on `dev` is the evidence that the arguments are rejected; the verifier that the removal is sufficient is reprobuild's `Test` job, which reaches this recipe now that metacraft-labs/reprobuild#355 has landed.

PR #764, opened 2026-09-21
ensure-ct-mcr asked the recorder for a target it does not export

zah:
`ensure-ct-mcr` runs `repro build --tool-provisioning=nix ct-mcr` in the `codetracer-native-recorder` sibling. There is no such target, and reprobuild's own diagnostic names the right one:

```
repro build: error: unknown_target: no build target matches 'ct-mcr'
repro build: did you mean:
  default
  codetracer-native-recorder.ct-cli
  codetracer-native-recorder.interpose
```

The recorder is not at fault. Its `repro.nim` declares `executable ctCli` with `name: "ct_cli"` and says why:

> INVOKED as `ct-mcr`, but the artifact the Justfile builds and signs is `ct_cli/ct_cli` ("The binary is placed at ct_cli/ct_cli (invoked as ct-mcr)"), so `name:` pins the on-disk spelling — calling it `ct-mcr` here would make this declaration disagree with every path in the Justfile and the packaging scripts.

`ct-mcr` is the invocation name; `ct-cli` is the build target.

### Why this was invisible

`ensure-ct-mcr`'s first arm is `exit 0` when `CT_CODETRACER_NATIVE_RECORDER_SIBLING` is unset, so the `repro build` arm runs only where the sibling is present:

| run | sibling | step | outcome |
|---|---|---|---|
| `dev`, job `106427423681`, 17:04 | not detected | SKIP arm, 14 s | green |
| PR #763, job `106504081929`, 20:37 | "repo present, ct_cli not built" | `repro build` arm | **red** |

Green exactly when it does nothing. The red on #763 was read as that PR's fault before this was traced; it was not.

Note the recipe directly below, `ensure-ct-native-replay`, already carries this lesson in its header — it deliberately does *not* route through `repro build` because that sibling has no reprobuild project. Here the sibling does have one; only the name was wrong.

### Verification

Not run locally: this arm needs the recorder sibling and its Nix dev shell. The evidence is reprobuild's own suggestion list plus the recorder's declaration. It is verified by `reprobuild-linux-smoke` on any runner that has the sibling — the same condition that exposed the bug.


zah:
**Correcting this PR's first revision — `ct-cli` was also wrong, and for an instructive reason.**

I took `ct-cli` from the `unknown_target` suggestion list. That list is not a list of selectors: `renderUnknownTargetDiagnostic` prints raw **action ids** (plus target and export names), so `codetracer-native-recorder.ct-cli` is an identifier the resolver stores, not a string you can pass back to it. Asking for `ct-cli` produced the identical error with the identical suggestion list:

```
repro build: error: unknown_target: no build target matches 'ct-cli'
repro build: did you mean:
  default
  codetracer-native-recorder.ct-cli
  codetracer-native-recorder.interpose
```

Reading the recorder's `repro.nim` instead of its diagnostic: it declares shell actions with explicit ids (`codetracer-native-recorder.ct-cli`, `.interpose`, `.dwarf-paths`, `.interpose-static`, `.ct-mcr.windows`) and registers exactly **one** named collection — `collect("default", defaultActions)`. There is no bare target under either spelling.

The selector that reaches one action inside a collection is `<collection>#<member>`. reprobuild's resolver suffix-matches `"." & memberName` against the collection's action ids; `.ct-cli` matches `codetracer-native-recorder.ct-cli` uniquely, so **`default#ct-cli`** resolves. That is what this PR now passes.

Amended rather than stacked, so the branch carries one correct commit. `default#ct-cli` is still unverified locally for the same reason as before — the arm needs the recorder sibling and its Nix dev shell — but it is now derived from the resolver's matching rule and the recorder's declarations rather than from a display string.

